PR #102, opened 2015-01-09
bugfix : app crash if a mail has 2 messageID

aenario:


aenario:
Damn, Brice has a terrible influence. Fixed, thanks !

PR #156, opened 2015-02-05
Switch to cozydb

aenario:
And some server-side fix


aenario:
rebased on 23dbb2c

aenario:
@clochix validated via Skype

PR #176, opened 2015-02-10
Improve button styles for compose and list widgets

frankrousseau:
I was annoyed by the fact buttons didn't change when I clicked on it. So, I changed it and made some minor improvements too.

* make them a litte bit bigger
* change color on click (active style)
* remove grey background on cancel button
* remove duplicated dependencies


clochix:
I merged it into #178, so I close this. Thanks Frank !

Issue #154, opened 2015-02-05
Account refreshing


poupotte:
When there's a problem refreshing an account (server temporally unavailable), refresh won't update next accounts.

poupotte:
Reopen it due to #163.

poupotte:
I can't reproduce this bug so I close this issue. If someone have this problem, don't hesitate to reopen it.

Issue #134, opened 2015-01-30
Problem to mark unread a specific email


poupotte:
I try to unread a message (with an attachment and open) and I receive an error. I have this message in my console :  
```
PATCH https://bellot-zoe.cozycloud.cc/apps/emails/message/883f98d550b30fe64ed500538313c6ff [HTTP/1.1 500 Internal Server Error 264ms]
"Error in messagePatch" "883f98d550b30fe64ed500538313c6ff" Object { message: "Invalid system flag \NONJUNK", stack: "Error: Invalid system flag \NONJUNK
    at Connection._resTagged (/usr/local/cozy/apps/emails/node_modules/imap/lib/Connection.js:1463:11)
    at Parser.<anonymous> (/usr/local/cozy/apps/emails/node_modules/imap/lib/Connection.js:173:10)
    at Parser.emit (events.js:95:17)
    at Parser._resTagged (/usr/local/cozy/apps/emails/node_modules/imap/lib/Parser.js:174:10)
    at Parser._parse (/usr/local/cozy/apps/emails/node_modules/imap/lib/Parser.js:138:16)
    at Parser._tryread (/usr/local/cozy/apps/emails/node_modules/imap/lib/Parser.js:82:15)
    at CleartextStream.Parser._cbReadable (/usr/local/cozy/apps/emails/node_modules/imap/lib/Parser.js:53:12)
    at CleartextStream.emit (events.js:92:17)
    at emitReadable_ (_stream_readable.js:427:10)
    at _stream_readable.js:420:7", source: "protocol" }
```
According to @clochix , NONJUNK tag, it is a tag added by thunderbird.

P.S. : Mark unread other mails works.

clochix:
AFAIK, JUNK and NONJUNK are flags used by Thunderbird (among others) but not supported by every email server.
When TB thinks a message is spam, it adds the Junk flag. If the user marks the message as not spam, TB then use the NoJunk flag to remember user's decision.
I checked and OVH server doesn't seem to support both of this flags.

aenario:
This is a server-side problem and is due to the distinction between server-supported flags and user-custom keywords. 

It should have been fixed by https://github.com/cozy/cozy-emails/commit/0f2a802174970b4186656b584a0e80153924813e , It is probably an error with the patch handling code.  @poupotte , if you can reproduce easily, could you get the request body ?

poupotte:
Request body : 
```
[...]

Issue #77, opened 2014-12-11
Delete conversation doesn't seem to work


clochix:

poupotte:
When I try to delete a conversation, I see it in INBOX and in trash folder.

poupotte:
It works fine with #181

poupotte:
Thanks @aenario

PR #181, opened 2015-02-13
Switch to cozydb

aenario:
Also include #134 #154 #77


clochix:
Thanks @aenario :-)
@frankrousseau I probably won't find time to review this this week-end. Would you mind have a look at it so we can merge ASAP ? Thanks !

frankrousseau:
Ok @clochix , I'll do it.

client/app/components/toolbox_actions.coffee
```diff
@@ -56,7 +54,7 @@ module.exports = ToolboxActions = React.createClass
                 if @props.message?
                     li role: 'presentation',
                         a
-                            href: "raw/#{@props.mailboxID}/#{messageRawId}"
+                            href: "raw/#{@props.message.get('id')}"
```
frankrousseau: Is that related to CozyDB. Are you sure about what you are doing?

server/config.coffee
```diff
@@ -34,7 +34,7 @@ config =
     ]
 
     plugins: [
-        MODEL_MODULE
+        'cozydb'
```
frankrousseau: Does it keep the pouchb compatibility?

server/controllers/messages.coffee
```diff
@@ -185,12 +186,17 @@ contentToBuffer = (req, attachment, callback) ->
...
-            return callback err if err
-            callback null, Buffer.concat buffers
+        chunks = []
+        bufferer = new stream.Writable
+        bufferer._write = (chunk, enc, next) ->
+            chunks.push(chunk)
+            next()
+        bufferer.end = ->
+            callback null, Buffer.concat chunks
+        fileStream.pipe bufferer
```
frankrousseau: Why do you need that? Does it mean you will keep the full file in memory?

server/controllers/messages.coffee
```diff
@@ -350,23 +358,43 @@ module.exports.send = (req, res, next) ->
     async.series steps, (err) ->
         return next err if err
         return next new Error('Server error') unless jdbMessage
-        res.send 200, jdbMessage.toClientObject()
+        out = jdbMessage.toClientObject()
+        out.isDraft = isDraft
+        res.send 200, out
```
frankrousseau: Use the new Express API:

```coffeescript
res.status(200).send out
```

By the way, if it's a 200 code, you don't need to mention it.

server/controllers/messages.coffee
```diff
@@ -350,23 +358,43 @@ module.exports.send = (req, res, next) ->
...
 
 
 module.exports.fetchConversation = (req, res, next) ->
```
frankrousseau: Comments?

server/controllers/messages.coffee
```diff
@@ -350,23 +358,43 @@ module.exports.send = (req, res, next) ->
...
+            next NotFound "conversation##{conversationID}"
+        else
+            log.debug "conversation##{conversationID}", messages.length
+            req.conversation = messages
+            next()
 
 module.exports.conversationGet = (req, res, next) ->
     res.send 200, req.conversation.map (msg) -> msg.toClientObject()
 
 module.exports.conversationDelete = (req, res, next) ->
```
frankrousseau: Comments?

server/controllers/messages.coffee
```diff
@@ -350,23 +358,43 @@ module.exports.send = (req, res, next) ->
...
+    accountID = req.conversation[0].accountID
+
+    Account.find accountID, (err, account) ->
+        return next err if err
+        return next new NotFound "Account##{accountID}" unless account
 
-    # @TODO : Delete Conversation
-    res.send 200, []
+        console.log account
```
frankrousseau: Dirty logging

server/imap/connection.coffee
```diff
@@ -204,6 +203,13 @@ Imap::multicopy = (uid, paths, callback) ->
         @copy uid, path, cb
     , callback
 
+# remove message from multiple boxes
+Imap::multiremove = (paths, callback) ->
+    async.eachSeries paths, ({path, uid}, cb) =>
+        @deleteMessageInBox path, uid, cb
```
frankrousseau: Is that possible to make a bulk delete with IMAP?

server/models/message.coffee
```diff
@@ -541,6 +560,7 @@ Message.createFromImapMessage = (mail, box, uid, callback) ->
 
     # we store normalized versions of subject & messageID for threading
     messageID = mail.headers['message-id']
+    delete mail.messageId
```
frankrousseau: I read that it's more performant to do that:

```coffeescript
mail.mesesageId = undefined
```

server/models/message.coffee
```diff
@@ -571,6 +592,8 @@ Message.createFromImapMessage = (mail, box, uid, callback) ->
                 name: att.generatedFileName
                 buffer: buffer
 
+        delete mail.attachments
```
frankrousseau: I read that it's more performant to do that:

```coffeescript
mail.mesesageId = undefined
```

client/app/components/toolbox_actions.coffee
aenario: This is not related to cozydb.

It is a seprate bugfix :  in conversation you could not see some raw messages.
Exemple : you have message A in inbox and message B in sent. You display the conversation from message A and it shows message B, but the message B is not in inbox so the messageRawID was undefined.

server/config.coffee
aenario: the pouchdb compatibility should be provided by cozydb (we discussed it in another PR, it still needs to be tested)

server/controllers/messages.coffee
aenario: We kept the full file in memory before and still do. We need to build the email to send it through SMTP & IMAP. Node-imap support for streams2 is not very good so until we can spend a few days making it solid, it is safer to buffer everything. 

The reason for this change is that jdb-cozy-adapter returned the mikeal/request object which act as a full stream (.on 'data') whereas cozydb return a hackish "LaterStream" which only has the .pipe and .abort method. I figure it was simpler to make this hack and remove it when we do the full streaming mentioned in the first § 

It will be cleaner to have it as a "Bufferer" class in helpers. Do you want me to make this change ?

server/controllers/messages.coffee
aenario: will do, thanks.

server/config.coffee
frankrousseau: What I mean is that, the MODEL_MODULE var ensured that pouch was used in case of standalone applicaiton. So, is it something removed or cozydb handles it itself?

server/imap/connection.coffee
aenario: It is possible to remove multiple message from a mailbox, but not as we are doing here a message that is in multiple mailboxes.

server/controllers/messages.coffee
frankrousseau: Don't waste too much time on it, just makes sure it works and explain everything through comments.

server/imap/connection.coffee
frankrousseau: ok

server/models/message.coffee
aenario: var a = {}
Object.keys(a) ---> Array [  ]
a.test = 'value'
Object.keys(a) ---> Array [ "test" ]
a.test = undefined
Object.keys(a) ---> Array [ "test" ]
delete a.test
Object.keys(a)   --->  Array [  ]
It is faster for "hot" objects (it prevents some interpreter optimization with hidden class), but its not exactly the same. I'd rather not risk introducing subtle bugs for a µs improvement of a once-called route.

server/config.coffee
aenario: cozydb handle it itself : if there is a db or dbName in americano options, it switch to pouchdb.
This should cover the cozy-light case. For the standole version to work, we need to set db or dbName from environement variable just above (https://github.com/aenario/cozy-emails/blob/d7efc8779304fb1abeb29d50b0bb9d2661018917/server.coffee#L15)

server/models/message.coffee
frankrousseau: Ok.

server/config.coffee
frankrousseau: ok

frankrousseau:
@clochix @aenario Can we merge it?

aenario:
i think so

aenario:
Let's merge it and fix bug in the next 3 days.

Issue #183, opened 2015-02-17
Remove progress bars during refresh


jsilvestre:
@frankrousseau and I discussed the fact that the orange progress bars are misleading for the users during the refresh operation.

I am going to remove them, and use spinner.svg to discreetly mark something is running in the background.

Let me know if you disagree / want something else.

poupotte:
Thanks @jsilvestre  for this suggestion, I'm ok  with you.

poupotte:
It works well, thanks @jsilvestre . It will be even better with new spinner :) !

jsilvestre:
I was not sure to add the new spinner in place of the existing button, that's why I didn't add it...

frankrousseau:
What did you do finally? From what I read, there is more progress bar and you finally drop the idea to use a spinner.

jsilvestre:
I removed the progress bar during refresh, and messed up my commit description (noticed it when it was merged). I didn't add the spinner, I used .fa-spin on the refresh icon instead, that looked more logical to me.

frankrousseau:
Ok. It makes things clearer. About the spinner, we want to use the same spinner in every place. That why we shouldn't use the Font awesome spinner.

frankrousseau:
The progress bar is still displayed on subfolders.

jsilvestre:
I don't use the font awesome spinner, .fa-spin makes spin whatever icon there is. It's the refresh icon that spins. That seemed more logical in the current UI, thus I agree that it's not coherent with other Cozy applications. Do you want me to the Cozy spinner anyway?

About the progressbar for subfolder, I will have a look.

frankrousseau:
Yes, I think the Cozy spinner is better.

jsilvestre:
Let's do it then.

PR #192, opened 2015-02-18
Fix previous progressbar

jsilvestre:
Based on feedback of my previous PR (see https://github.com/cozy/cozy-emails/issues/183#issuecomment-74861001):
* changed: the spinner for Cozy's official one
* fixed: the progress bar for mailboxes (dunno how I missed it at first)

I also noticed a localization was missing.

### DIFF
--- a/client/app/components/message-list.coffee
+++ b/client/app/components/message-list.coffee
@@ -200,16 +200,18 @@ MessageList = React.createClass
                         # refresh
                         if not @state.edited
                             div className: btnClasses,
-                            refreshClasses = 'fa fa-refresh'
-                            if @props.refreshes.length > 0
-                                refreshClasses += ' fa-spin'
-                            div className: 'btn-group btn-group-sm message-list-option',
-                                button
-                                    className: 'btn btn-default',
-                                    type: 'button',
-                                    disabled: null,
-                                    onClick: @refresh,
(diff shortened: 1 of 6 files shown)


client/app/components/message-list.coffee
```diff
@@ -200,16 +200,18 @@ MessageList = React.createClass
...
+                                        className: 'btn btn-default',
+                                        type: 'button',
+                                        disabled: null,
+                                        onClick: @refresh,
+                                            span className: 'fa fa-refresh'
+                                else
+                                    img
+                                        src: 'images/spinner.svg'
+                                        alt: 'spinner'
+                                        role: 'spinner'
```
m4dz: As `[role=spinner]` doesn't exist in ARIA spec, I think it won't bring anything here and can make things more confuse.

If we cannot handle the throbbers in pseudo-elements, then prefer a class here to target it with CSS. Even if custom roles are allowed by the specs, a throbber should not be in the markup as it's only a state indicator, so a class doesn't hurt me as well :wink:.

jsilvestre: Fixed. This whole thing is almost a 100% BS to me. I definitely need a training!

m4dz: With an upcoming guideline, I promise :smile:!

PR #202, opened 2015-02-25
Refactoring of client side code

clochix:
There should be no visible effect.


clochix:
Ping @aenario

client/app/components/account-config.coffee
```diff
@@ -688,8 +688,13 @@ AccountConfigMailboxes = React.createClass
 
     render: ->
         favorites = @state.favoriteMailboxes.value
+        mailboxesFlat = {}
         if @state.mailboxes.value isnt '' and favorites isnt ''
             mailboxes = @state.mailboxes.value.map (mailbox, key) =>
+                id = mailbox.get 'id'
+                mailboxesFlat[id] = {}
```
aenario: move this to an helper function

client/app/components/account-config.coffee
```diff
@@ -771,6 +776,13 @@ AccountConfigMailboxes = React.createClass
 
     _renderMailboxChoice: (labelText, box) ->
         if @state.id? and @state.mailboxes.value isnt ''
+            mailboxesFlat = {}
+            @state.mailboxes.value.map (mailbox) ->
+                id = mailbox.get 'id'
+                mailboxesFlat[id] = {}
```
aenario: move this to an helper function

aenario: Or may be to the state itself ? This way this part can stay unmutated. 
Why is selectedID copied to each mailbox ?

client/app/components/application.coffee
```diff
@@ -406,13 +414,33 @@ module.exports = Application = React.createClass
         else
             selectedMailboxID = null
 
+        accounts = AccountStore.getAll()
+        accountsFlat = {}
```
aenario: idem

client/app/components/mails_input.coffee
```diff
@@ -159,7 +159,7 @@ module.exports = MailsInput = React.createClass
                 "#{contact.get 'fn'} <#{contact.get 'address'}>"
 
     onQuery: (char) ->
-        query = @refs.contactInput.getDOMNode().value.split(',').pop().trim()
+        query = @refs.contactInput.getDOMNode().value.split(',').pop().trimLeft()
```
aenario: is this standard ?

server/models/requests.coffee
```diff
@@ -10,7 +10,7 @@ module.exports =
 
     contact:
         all: cozydb.defaultRequests.all
-        byName: (doc) ->
+        mailByName: (doc) ->
```
aenario: DS should handles conflict resolution. Not sure if we should also prefix in each app for additional safety in dev mode. @poupotte @frankrousseau , any opinion ?

aenario:
Please add some comments regarding the flattening + de-immutabilisation of parameters.  

I am not sure if your request name changes on server-side are necessary / a good idea. May be we just want to enable the DS request conflict resolution in NODE_ENV=development...

client/app/components/mails_input.coffee
clochix: No, you're right, it works in Firefox and Chromium but may fail in other engines, I'm gonna fix it.

PR #221, opened 2015-03-20
Improve actions on conversations

clochix:


clochix:
I totally messed up my latest pull and now my history is totally f****ed. I will delete that **** and try a new PR some day ☹

PR #234, opened 2015-03-26
Fix #215 + minor potential bug

### LINKED ISSUE #215: Nothing happens when refreshing if login/pass is incorrect
If your login, or more often your password becomes incorrect after a successful configuration, nothing seems to happen when you try to synchronize the mailbox.
-> A click on the refresh icons launches nothing visually (the spinner is not showing) and no error message is displayed.

If you come back to the mailbox config, the errors on address / password are well highlighted.

After you change your password for the good one, come back to the email view and try to synchronize again, everything goes smooth, i.e the spinner is spinning on click and your mails are fetched.

jsilvestre:
* the timeout for automatic refresh is not cleared, which could result in multiple refreshes started at the same time.
* fix #215

### DIFF
--- a/server/models/account.coffee
+++ b/server/models/account.coffee
@@ -43,6 +43,7 @@ log = require('../utils/logging')(prefix: 'models:account')
 _ = require 'lodash'
 async = require 'async'
 CONSTANTS = require '../utils/constants'
+{RefreshError} = require '../utils/errors'
 require('../utils/socket_handler').wrapModel Account, 'account'
 
 refreshTimeout = null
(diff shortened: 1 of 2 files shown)


aenario:
Today I learned there is no need to check if a timeout exist before clearing it. Thanks @jsilvestre

jsilvestre:
I was not sure of myself, but I did it before and it never threw. Did you look for/find a source?

aenario:
nope, I opened up a console and tried to clearTimeout null and clearTimeout expiredTimeout and it did not threw.

PR #237, opened 2015-03-26
Merge early, merge often

clochix:


client/app/actions/conversation_action_creator.coffee
```diff
@@ -13,13 +13,21 @@ module.exports =
         conversation = MessageStore.getConversation conversationID
         account = AccountStore.getByID(conversation.get(0).get 'accountID')
         trash   = account.get 'trashMailbox'
+        if not trash? or trash is ''
+            callback t('message delete no trash')
```
aenario: Can you callback with an Error object instead of a string ?

client/app/actions/conversation_action_creator.coffee
```diff
@@ -32,6 +40,10 @@ module.exports =
 
         .toJS()
 
+        if error isnt ''
+            callback error
```
aenario: same

client/app/actions/message_action_creator.coffee
```diff
@@ -173,6 +173,8 @@ module.exports =
 
     # set conv to true to update current conversation ID
     setCurrent: (messageID, conv) ->
+        if typeof messageID isnt 'string'
```
aenario: cant you fix instead where it is passed as Message instead of MessageID ?

aenario:
Discussed on IRC : we dont use error elsewhere in the client & its better to be safe.

PR #245, opened 2015-03-27
Fix callbacks passing to toolbar / toolbox message actions

m4dz:
This PR contains a bug when marking a message. WIP.


m4dz:
Ok for me :)

PR #254, opened 2015-03-30
Misc fixes

clochix:


client/app/components/message-list.coffee
```diff
@@ -591,16 +591,17 @@ MessageItem = React.createClass
         ,
             tag
                 href: url,
-                className: 'wrapper'
+                'data-href': url,
```
aenario: why do you need both href & data-href ?

aenario:
Could you rm -rf build/client & rebuild ? There is some artefacts from squire.

PR #258, opened 2015-03-31
Handle better unknown contacts

frankrousseau:
* Allow the user to enter a contact by typing enter.
* Display an error message if the email is not properly formatted.


frankrousseau:
I added a modification to display the stylish spinner instead of the font awesome one.
It brings a new files spinner-white.svg. Maybe there is smarter way than adding a new icon. Not sure about the best way to do.

client/app/utils/message_utils.coffee
```diff
@@ -117,8 +123,21 @@ module.exports = MessageUtils =
...
 
             when ComposeActions.FORWARD
-                separator = t('compose forward separator',
-                    {date: dateHuman, sender: sender})
+                console.log inReplyTo.get 'to'
+                addresses = ''
+                for address in inReplyTo.get 'to'
+                    addresses += "#{address.address}, "
+                if addresses.length > 3
+                    addresses = addresses.substring 0, addresses.length - 2
```
aenario: it would be cleaner to use a map & join : 
```
inReplyTo.get('to')
   .map (address) -> address.address
   .join ', '
```

aenario: and remove the console.log

frankrousseau: done

Issue #263, opened 2015-04-01
Ui: Conversation view


m4dz:
_Note: This issue resume #255_

# Revamp
* Conversation view
    * improve semantics by isolating components in markup (section, article, header, nav…)
    * add a toolbar component that handles _large view_, _prev_/_next_ navigation
    * isolate conversation subject to top header
* Messages stack in conversation
    * messages are now stacked with a compact header (sender infos, date and indicators)
    * if the stack length is > 3, then only load first and last message from the stack and add a button to load the full stack (in compact mode)
    * a click on compact header will toggle the message view (compact or full)
* Message view
    * header display only sender and recipients
    * _from_ label is removed
    * indicators are only displayed in header (no more links)
    * details box is available to display more header informations
    * toolbar is now right-aligned and buttons are correctly marked / grouped
    * all buttons in toolbar now uses dedicated icons
    * a subbar in footer is available for quick-reply actions (a quick reply compose box can be a nice improvement)
    * attachments are now displayed in footer with a visualisation link and a download link
    * attached images are displayed as thumbnail in footer (a real thumbnail generator can be a nice improvement)

# Global Style
* toolbars use the bootstrap style
* messages stack and message view now use a card-style
* `pre` markup uses the given _Source Code Pro_ font

# Known bugs and temporary disabled features
* contact tooltip (_Participant_ component) is temporary disabled because it needs to be revamped, with preview box and quick-add features
* the message iframe seems to be too high
* the large messge view switch is currently broken, it will be fixed in another PR

poupotte:
It's very nice :) 
Good job, I'm really surprised of result !

Some remarks/questions : 
* When a contact hasn't photo (because is unknown for exemple), it's displayed in black instead of blue (in message stack conversation), is it wanted? Maybe it's due to disable of contact tooltip.
* I would add a line 'Move to...' at the beginning of menu (folder button).
* There isn't tooltips on button, it's for an other PR ?
* Attachments seems available only at the end of mail. Why not use paper clip as a button ?

m4dz:
> Good job, I'm really surprised of result !

Thanks a lot!

> Some remarks/questions :
> * When a contact hasn't photo (because is unknown for exemple), it's displayed in black instead of blue (in message stack conversation), is it wanted? Maybe it's due to disable of contact tooltip.

Yep, that's it. I'll produce quickly the component, probably with @Benibur whose want to contribute to this part.

> * I would add a line 'Move to...' at the beginning of menu (folder button).

[...]

PR #270, opened 2015-04-02
Ui: Conversation view fixes

m4dz:
This PR contains various fixes mentioned in #263 comments.


client/app/components/attachement_preview.coffee
```diff
@@ -0,0 +1,56 @@
...
+    render: ->
+        if @props.previewLink
+             li key: @props.key,
+                @renderIcon()
+                a
+                    target: '_blank'
+                    href: @props.file.url,
+                        # TODO: generate a thumb instead of loading raw file
+                        img width: 90, src: @props.file.url if @props.preview
+                        @props.file.generatedFileName
```
m4dz: Hummm… unfortunately, it seems that the DS generate thumb on for _File_ documents, and attachment aren't stored as _File_. It can't get a valid `thumb` key which contains the thumb at this time. I'll suggest we deal with this improvement later :).

frankrousseau: ok

client/app/components/attachement_preview.coffee
```diff
@@ -0,0 +1,56 @@
...
+
+
+    renderIcon: ->
+        type = MessageUtils.getAttachmentType @props.file.contentType
+        i className: "fa #{@icons[type] or 'fa-file-o'}"
+
+
+    displayFilesize: (length) ->
+        if length < 1024
+            "#{length} octets"
```
aenario: This need translating

client/app/components/message_header.coffee
```diff
@@ -112,4 +144,7 @@ module.exports = React.createClass
                         row @props.message.get('subject'), 'headers subject'
 
     toggleDetails: ->
-        @setState detailled: !@state.detailled
+        @setState showDetails: !@state.showDetails
```
aenario: use `not` rather than `!`

Issue #149, opened 2015-02-03
Improve spinner


poupotte:
I suggest to use cozy-mobile spinner   https://github.com/poupotte/cozy-files-mobile/blob/master/www-src/app/assets/img/spinner.svg for email application.

frankrousseau:
See PR: https://github.com/cozy/cozy-emails/pull/258

frankrousseau:
Done.

poupotte:
Thanks @frankrousseau

PR #279, opened 2015-04-02
Fix spinner size/position and email validation behavior

frankrousseau:
* Related to requirements asked by @poupotte on #149 (size)
* By the way I improved the compose button styles a little bit.
* Fix the bugs that make the validation error popup pop too often (#258)


frankrousseau:
@aenario @clochix  Because it's the end of the sprint and because it's not so much modification, I merge this pull request myself.

PR #282, opened 2015-04-06
Improve conversation style

frankrousseau:
I propose here two style modifications. @m4dz tell me if you think it's wrong.

* change mail title to Source Sans
* make the image warning matches more the Cozy style


m4dz:
Thank @frankrousseau :)

PR #283, opened 2015-04-07
Fix contacts drag'n'drop

clochix:


frankrousseau:
Can we merge it or do you prefer to handle it with JQuery?

clochix:
It works now, so let's merge it

Issue #289, opened 2015-04-07
Should not add twice the prefix Re: in the reply subject


frankrousseau:
When I reply to an email I don't want to duplicate the "Re:" prefix in the subject.

poupotte:
It's ok for me. Thanks @frankrousseau

Issue #287, opened 2015-04-07
Add mail in header of forwarded messages


poupotte:

poupotte:
It works for me. Thanks @frankrousseau

Issue #286, opened 2015-04-07
Remove line breaks on the top of reply mail


poupotte:

poupotte:
It works well, thanks @frankrousseau

PR #291, opened 2015-04-08
Minor fixes

frankrousseau:
* Do not add twice reply prefix in reply subject
* Remove blank lines at the beginning of the reply message 	
* Properly build from field in forward message header

Ref: #286, #287, #289


aenario:
Thanks !

PR #302, opened 2015-04-13
Improve account counter (ignore junk & trash)

aenario:
and faster synchro
some linting & comments

### DIFF
--- a/server/models/mailbox.coffee
+++ b/server/models/mailbox.coffee
@@ -82,7 +82,7 @@ class Mailbox extends cozydb.CozyModel
         Mailbox.getBoxes accountID, (err, boxes) ->
             return callback err if err
             boxIndex = {}
-            boxIndex[box.id] = path: box.path for box in boxes
+            boxIndex[box.id] = box for box in boxes
             callback null, boxIndex
 
     # Public: remove mailboxes linked to an account that doesn't exist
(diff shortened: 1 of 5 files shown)


server/models/mailbox.coffee
```diff
@@ -487,13 +487,13 @@ class Mailbox extends cozydb.CozyModel
...
-            Message.fetchOrUpdate box, msg.mid, msg.uid, (err, result) ->
+            Message.fetchOrUpdate box, msg, (err, result) ->
                 reporter.onError err if err
                 reporter.addProgress 1
                 if result?.shouldNotif is true
                     shouldNotif = true
-                # dont stop
-                cb null
+                # loop anyway, let the DS breath
+                setTimeout (-> cb null), 50
```
frankrousseau: Isn't it the same to run `cb()` (instead of `cb null`)?

frankrousseau:
Thank you @aenario

frankrousseau:
The counter works well for my inbox. But the overall counter is still broken.

frankrousseau:
I think we should simply display no counter on the mail account level.

server/models/mailbox.coffee
aenario: it is. Will fix, thanks !

Issue #277, opened 2015-04-02
Download an attachment with space in name doesn't work well


poupotte:
For example, if I try to download `my super document.txt`, my downloaded file is named `my`.

aenario:
Should have been fixed in https://github.com/cozy/cozy-emails/commit/0dd1cf4d11c54b0cb8771e6c1ce07567bc181c7e

poupotte:
Seems broken : I've an attachment with spaces, I can open it in another window but if I try to download it, I receive an error  : 

```
The connection to wss://bellot-zoe.cozycloud.cc/apps/emails/socket.io/?EIO=3&transport=websocket&sid=FWD4BTCthh1thGRiAAAC was interrupted while the page was loading. vendor.js:60598:0
GET https://bellot-zoe.cozycloud.cc/apps/emails/message%2Fa50d6ec3154960c7619612932e039c42%2Fattachments%2FBIA%20PREV-%C3%A0%20remplir%20par%20chaque%20salari%C3%A9.pdf [HTTP/1.1 404 Not Found 15ms]
```

poupotte:
It works well, thanks @jsilvestre

PR #306, opened 2015-04-13
Fixed: dowloading attachements resulted in 404. Refs #277

jsilvestre:
* fixes https://github.com/cozy/cozy-emails/issues/277
* I have no idea why commit c47b250 is here...I chose not to mess around with git to prevent losing data.


frankrousseau:
Thank you!

Issue #138, opened 2015-02-02
Add tooltips for action button


poupotte:

poupotte:
Mmmm... it works on the first message, but if I open older messages in a conversation, tooltips don't work for its.

poupotte:
Moreover, when I click on a button after tootip apparition, tooltip doesn't disappear,  I have to move my mouse.
![tooltip](https://cloud.githubusercontent.com/assets/2659119/7131463/f8c2a27c-e27f-11e4-99b9-b28fc3c5dff6.png)

jsilvestre:
About the tooltip that doesn't fadeout, I opened an issue on aria-tips: https://github.com/m4dz/aria-tips/issues/4.

frankrousseau:
I face the same bugs.

jsilvestre:
I managed to reproduce the bug that on other messages, tooltips are not shown. I'm just scratching a bit my head around the matter. Also pushed fixes to aria-tips, I will work a bit more on it in order to make it smoother.

jsilvestre:
@poupotte I did more improvements, there is still the bug you mentioned because I don't manage to reproduce it with aria-tips alone...What should we do with this issue?

poupotte:
Issue in aria-tips is sufficient for the issue but it lacks tooltips for some buttons, so this issue must stay opened.

poupotte:
Thanks @jsilvestre

PR #308, opened 2015-04-13
Tooltips on action conversation buttons

jsilvestre:
* Added structure to manage tooltips with aria-tips lib
* Added tooltips (see #138)
* Also very minor style change that was annoying me: the outline on the button to expand hidden messages in a conversation.


frankrousseau:
@jsilvestre did you see with  @m4dz before doing it? I mentioned a lib he built to do it.

frankrousseau:
Whatever this PR lacks of comments

jsilvestre:
aria-tips is @m4dz's lib. I'm adding comments.

jsilvestre:
I added more comments and built the code.

client/app/locales/en.coffee
```diff
@@ -359,3 +359,14 @@ module.exports =
       'calendar unknown format' : """
             This message contains an invite to an event in a currently unknown format.
             """
+
+      # Tooltips
+      "tooltip reply": "Answer"
+      "tooltip reply all": "Answer to all"
+      "tooltip forward": "Forward"
+      "tooltip remove message": "Remove"
+      "tooltip open attachments": "Open attachments list"
```
frankrousseau: There is no `s` at `attachment`. This word has an epithet value.

jsilvestre: Weird, so you say "attachment list", but you also say "a list of attachments"?

frankrousseau: Yes, I think it works that way.

PR #310, opened 2015-04-14
Display all info and error alerts through toasters

frankrousseau:
To unify alert messages, this PR make Emails always displaying a toaster when something happens. It simplifies code and UI behavior. 
By the way I splitted the toaster file in two: one for each components.


aenario:
Minor fix : https://github.com/cozy/cozy-emails/commit/b1c0cb6a5c5f90cc66e4f184fcdf3ba776c81eb2
Thanks !

PR #349, opened 2015-04-17
Few tooltip related fixes

jsilvestre:
I did some fixes/improvements around the attachment list visible in the message itself (not at the bottom, when you click on the icon).


frankrousseau:
Thank you!

PR #351, opened 2015-04-17
Improve Details Header

m4dz:
This commit introduces few changes and improvements:
* Externalize participants formatting in a mixin for later use by
Participant component
* Add click outside mixin for modals / popups components
* Add the clickoutside handler on Details message header popup to close
the popup

### DIFF
--- a/client/app/components/message_header.coffee
+++ b/client/app/components/message_header.coffee
@@ -2,14 +2,19 @@
 MessageUtils = require '../utils/message_utils'
 
 AttachmentPreview = require './attachement_preview'
+PopupMessageDetails = require './popup_message_details'
 
-ContactStore   = require '../stores/contact_store'
+ParticipantMixin = require '../mixins/participant_mixin'
 
 {MessageFlags, Tooltips} = require '../constants/app_constants'
 
 
 module.exports = React.createClass
     displayName: 'MessageHeader'
+
+    mixins: [
+        ParticipantMixin
+    ]
 
     propTypes:
         message: React.PropTypes.object.isRequired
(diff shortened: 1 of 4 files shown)


aenario:
Merged in development. 
Thanks @m4dz

frankrousseau:
I noticed two problems with that PR:

* Lack of comments
* Usage of function instead of React component for the row component.

m4dz:
I agree for the lack of comments, I'll fix this in a next PR about popups :wink:.

About the row function, I'm not sure a dedicated React component can really be usefull here as it's just a convenient wrapper to build table rows, which can be done directly in the main rendering thread but will exceed 80 chars per line with this f*** amrkup indent. But I can be wrong… @jsilvestre, any recommendation here?

jsilvestre:
I see two issues with the current implementation choice:
* the function should be attached to "this" so the `render` function is as easy and small as possible.
* it should be named "getXXXXRender" because using "row" it makes people think you are using a component, whereas you are not (I got caught myself in this piece of code, looking for the `row` component)

This is my personal vision of having sub render function in a React component, I have never seen a guideline or good practice article about it.

If you want to discuss the implementation choice itself (sub render function vs component), as I said the other day it is hard to define a rule, because in some case the sub render function is better, in some others the component is better.
My personal rule is the following: if all you need is a wrapper around a single html tag, go for a sub render function, otherwise make a component. That specific case is right on the limit of this rule, thus is interesting. If you apply the rule, you must go for a separate component. The problem is that it's obviously is a limit case, with a bit a rewriting to make it more explicit, the `row` function could be used as-is (see also my first two remarks) and be easier to work with than an external file.

In the end, I'd still go for the external file to match my rule, because even if there are limits, coherence is still the best way to go. Then again, it's *my* rule, not Cozy's.

Issue #357, opened 2015-04-20
Account component refactoring 


frankrousseau:
I think we do not develop properly with React. That makes the application hard to maintain. Moreover we don't take advantage of the testability of React.

I didn't read much blog post about React and its siblings but I think we can apply basic principles of UI development. That includes:

* Proper naming of each component.
* Use a set of basic components reusable among main widgets (button, title, etc.).
* Move logic code that prepares the rendering into subfunctions.
* Have one file for each non basic components
* Add comments

Once done, I think we could think about the current architecture and see if we can remove logic from the component or if we should keep it that way. Currently the view is very smart compared to a pure MVC architecture, that could be a problem.

Finally we will be able to work on unit testing with the karmajs/mocha combo or equivalent.

To make this study, I will work on the account creation component.

frankrousseau:
I think the work is done. It's now about bugfixing. Things seem to be proper enough.

PR #359, opened 2015-04-20
AccountConfig: component cleaning

frankrousseau:
Please DO NOT MERGE

This pull request is related to #357. 

It's a work in progress. I think that most of the job is done, but it's broken, inputs can't be changed anymore. I would like we discuss it too.


frankrousseau:
I close it to change the target branch.

PR #362, opened 2015-04-20
AccountConfig component cleaning

frankrousseau:
This pull request is related to #357.

Main idea behind this PR is 

* to use more sub components
* have a set of basic components reusable among bigger components
* comment almost every methods
* have more functions but shorter functions.
* one file per component

Ready to merge!


client/app/components/account_config_item.coffee
```diff
@@ -0,0 +1,186 @@
...
+        RouterMixin
+        React.addons.LinkedStateMixin # two-way data binding
+    ]
+
+    propTypes:
+        mailbox: React.PropTypes.object
+
+
+   # Do not update component if nothing has changed.
+   #shouldComponentUpdate: (nextProps, nextState) ->
```
clochix: You forgot to remove comments here.

client/app/components/account_config.coffee
```diff
@@ -0,0 +1,382 @@
...
+        options =
+            isWaiting: @props.isWaiting
+            selectedAccount: @props.selectedAccount
+            validateForm: @validateForm
+            onSubmit: @onSubmit
+            onBlur: @onFieldBlurred
+            errors: @state.errors
+            checking: @state.checking
+
+        options[field] = @linkState field for field in @_accountFields
```
jsilvestre: I didn't see that being used in `AccountInput`, which uses the `LinkStateMixin` itself. Did I miss something or is there and old unused behaviour?

client/app/components/account_config.coffee
```diff
@@ -0,0 +1,382 @@
...
+
+    # Build options required by mailbox tab.
+    buildMailboxesOptions: (options) ->
+
+        options =
+            error: @props.error
+            errors: @state.errors
+            onSubmit: @onSubmit
+
+        options[field] = @linkState field for field in @_mailboxesFields
```
jsilvestre: Same remark here than above.

client/app/components/account_config.coffee
```diff
@@ -0,0 +1,382 @@
...
+                Tabs tabs: tabParams
+            if not @props.tab or @props.tab is 'account'
+                AccountConfigMain mainOptions
+            else
+                AccountConfigMailboxes mailboxesOptions
+
+
+    # Ask to main layout manager to display error as notification toasters.
+    renderError: (errors) ->
+        LayoutActionCreator.alertError t 'account errors'
```
jsilvestre: This function has no meaning with your refactoring, you could use `LayoutActionCreator.alertError` directly in the validation method. Or if you really want to keep it, rename it, so it doesn't confuse with an actual sub render function.

client/app/components/account_config_input.coffee
```diff
@@ -0,0 +1,97 @@
...
+        return mainClasses
+
+
+    # Build input placeholder depending on given type (build right translation
+    # key and run translation process on it).
+    buildPlaceHolder: (type, name) ->
+        placeHolder = null
+        if type is 'text' or type is 'email'
+            placeHolder = t("account #{name} short")
+        placeHolder
```
jsilvestre: Please, help me nevrosis and don't use implicit return :(

client/app/components/account_config_input.coffee
```diff
@@ -0,0 +1,97 @@
...
+    renderError: (errorField, name) ->
+        result = []
+        if Array.isArray errorField
+            for error in errorField
+                if @state.errors?[error]? and error is name
+                    result.push ErrorLine text: @state.errors[error]
+        else
+            if @state.errors?[errorField]?
+                result.push ErrorLine text: @state.errors[errorField]
+        result
```
jsilvestre: Please, help my nevrosis and don't use implicit return :(

client/app/components/account_config_input.coffee
```diff
@@ -0,0 +1,97 @@
...
+        React.addons.LinkedStateMixin # two-way data binding
+    ]
+
+
+    getInitialState: ->
+        return @props
+
+
+    componentWillReceiveProps: (props) ->
+        @setState props
```
jsilvestre: There are no point in turning the props into state in this component since it only reads the props. You can simply use `@props.` rather than `@state.` and remove `getInitialState`  and `componentWillReceiveProps`. Putting **some** props into the state should be done carefully and is considered an anti-pattern (see https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html).

jsilvestre:
Global remark: either we decide to use `propTypes` for all components or we don't, but it is used sometimes and sometimes it is not. Let's just take a decision and stick to it.

client/app/components/account_config_item.coffee
```diff
@@ -0,0 +1,179 @@
...
+        pusher = ""
+        pusher += "    " for j in [1..@props.mailbox.get('depth')] by 1
+        key = @props.mailbox.get 'id'
+
+        if @state.favorite
+            favoriteClass = "fa fa-eye mailbox-visi-yes"
+            favoriteTitle = t "mailbox title favorite"
+        else
+            favoriteClass = "fa fa-eye-slash mailbox-visi-no"
+            favoriteTitle = t "mailbox title not favorite"
```
jsilvestre: I'd put that in a sub function (like you did in other components), especially since it's far from where it is actually used.

client/app/components/account_config_item.coffee
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+
+    # Whether it's edit mode or not, it displays widgets to edit mailbox
+    # properties like the name. Otherwise it displays box information and
+    # button to switch to edit mode and button to delete the current mailbox.
+    render: ->
+        pusher = ""
+        pusher += "    " for j in [1..@props.mailbox.get('depth')] by 1
+        key = @props.mailbox.get 'id'
```
jsilvestre: This could be a sub function.

client/app/components/account_config_mailboxes.coffee
```diff
@@ -0,0 +1,210 @@
...
+                        mailboxes: @state.mailboxesFlat
+                        selectedMailboxID: @state[box].value
+                        onChange: (mailbox) =>
+                            # requestChange is asynchroneous, so we need
+                            # to also call setState to only call onSubmet
+                            # once state has really been updated
+                            @state[box].requestChange mailbox
+                            newState = {}
+                            newState[box] =
+                                value = mailbox
```
jsilvestre: I am not sure of what you meant here.

client/app/components/account_config_main.coffee
```diff
@@ -0,0 +1,471 @@
...
+
+AccountActionCreator = require '../actions/account_action_creator'
+AccountInput = require './account_config_input'
+
+RouterMixin = require '../mixins/router_mixin'
+LayoutActionCreator = require '../actions/layout_action_creator'
+{Form, FieldSet, FormButtons, FormDropdown} = require './basic_components'
+
+
+state = null
```
jsilvestre: Why do you use a variable in the file scope rather than the component's state?

client/app/components/account_config_main.coffee
```diff
@@ -0,0 +1,471 @@
...
+                errors: @state.errors
+                errorField: ['imap', 'imapServer', 'imapPort']
+                onBlur: @props.onBlur
+
+            AccountInput
+                name: 'imapPort'
+                value: @linkState('imapPort').value
+                errors: @state.errors
+                onBlur: =>
+                    @_onimapPort
```
jsilvestre: Is it actually called without the `()`?

client/app/components/account_config_main.coffee
```diff
@@ -0,0 +1,471 @@
...
+                ]
+                onBlur: @props.onBlur
+
+            AccountInput
+                name: 'smtpPort'
+                value: @linkState('smtpPort').value
+                errors: @state.errors
+                onBlur: =>
+                    @props.onBlur()
+                    @_onSMTPPort
```
jsilvestre: Is it actually called without the `()`?

client/app/components/account_config_main.coffee
```diff
@@ -0,0 +1,471 @@
...
+    # Do not update component if nothing has changed.
+    shouldComponentUpdate: (nextProps, nextState) ->
+        isNextState = _.isEqual nextState, @state
+        isNextProps = _.isEqual nextProps, @props
+        return not (isNextState and isNextProps)
+
+
+    _propsToState: (props) ->
+        state = props
+        return state
```
jsilvestre: All props are merged in to state, only props we want in state should be in the state to ease understanding of how the form works.

client/app/components/basic_components.coffee
```diff
@@ -0,0 +1,166 @@
...
+
+
+Container = React.createClass
+
+    render: ->
+        div
+            id: @props.id
+            key: @props.key
+        ,
+            @props.children
```
jsilvestre: Okay for some markup reuse, but this one is too much IMHO ;-)

client/app/components/account_config_mailboxes.coffee
```diff
@@ -0,0 +1,210 @@
...
+        if @state.mailboxes.value isnt '' and favorites isnt ''
+            mailboxes = @state.mailboxes.value.map (mailbox, key) =>
+                try
+                    favorite = favorites.get(mailbox.get('id'))?
+                    MailboxItem {accountID: @state.id.value, mailbox, favorite}
+                catch error
+                    console.error error, favorites
+            .toJS()
+
+        form className: 'form-horizontal',
```
jsilvestre: Why don't you reuse the Form basic component?

server/controllers/accounts.coffee
```diff
@@ -53,6 +53,7 @@ module.exports.create = (req, res, next) ->
 # check account parameters
 module.exports.check = (req, res, next) ->
     tmpAccount = new Account req.body
+    console.log tmpAccount
```
jsilvestre: Any reason for this?

jsilvestre:
Besides a few things, not necessarily related to your changes, it's looks good!

client/app/components/account_config.coffee
frankrousseau: I don't know, it's just a copy paste of the previous code. I don't really understand what does this line.

client/app/components/account_config.coffee
frankrousseau: I don't know, it's just a copy paste of the previous code. I don't really understand what does this line.

client/app/components/account_config.coffee
frankrousseau: ok

client/app/components/account_config_input.coffee
frankrousseau: Another option: cure your nevrosis?

client/app/components/account_config_input.coffee
frankrousseau: I didn't change anything to this function. It's just a copy/paste.

client/app/components/account_config_input.coffee
aenario: tumblr_inline_nmomtok4il1s9mb2r_540

jsilvestre: Well that was just a pretext, implicit returns are evil ;-)

client/app/components/account_config.coffee
jsilvestre: I think you can remove them.

client/app/components/basic_components.coffee
frankrousseau: Explicit component name is far more explicit than html tags.

client/app/components/account_config_mailboxes.coffee
frankrousseau: ok

server/controllers/accounts.coffee
frankrousseau: no

client/app/components/account_config_main.coffee
frankrousseau: It's a copy paste from the previous code.

client/app/components/account_config_main.coffee
frankrousseau: ok

client/app/components/account_config_main.coffee
frankrousseau: ok

client/app/components/account_config_main.coffee
frankrousseau: It's a copy paste from the previous code, I was afraid to break something by removing it.

client/app/components/account_config_mailboxes.coffee
frankrousseau: me neither :)

client/app/components/account_config_item.coffee
frankrousseau: ok

client/app/components/account_config_item.coffee
frankrousseau: ok

client/app/components/account_config_input.coffee
frankrousseau: You're right pretty things are often evil. But "we can't be evil". So I'll remove it.

client/app/components/account_config_input.coffee
frankrousseau: Please change that part later if you want. I don't know the application well enough to change that properly.

frankrousseau:
> Global remark: either we decide to use propTypes for all components or we don't, but it is used 
> sometimes and sometimes it is not. Let's just take a decision and stick to it.

Can you add an explaination of `propTypes` to the Cozy Emails wiki so we can apply it properly.

client/app/components/account_config.coffee
frankrousseau: When I remove it, it breaks some stuff in the component.

client/app/components/account_config.coffee
frankrousseau: When I remove it, it breaks some stuff in the component.

jsilvestre:
It's done, thank you. I merge.

client/app/components/account_config.coffee
jsilvestre: I'll have a look at it during quality time.

PR #367, opened 2015-04-22
[HOTFIX] Incompatibility with couchdb 1.3

aenario:
The new bymailboxrequest used Object.keys which is not available in couchdb 1.3 javascript environment.

I am making the PR as an "hotfix" against master, because this bug will cause all mails to be reimported twice for user using couchdb 1.3 ...


poupotte:
It's ok for me.

PR #375, opened 2015-04-24
Fix previous merge

frankrousseau:
Quickfix for formatUsers duplication problem:

What would be best: to move the formatUser version of message_header to the ParticipantMixin.


frankrousseau:
Wait before merging I introcuded a typo.

frankrousseau:
Ok it should work now.

PR #378, opened 2015-04-24
Improved: feedback during refresh

jsilvestre:


client/app/components/compose.coffee
```diff
@@ -191,10 +189,7 @@ module.exports = Compose = React.createClass
                                 disable: if @state.saving then true else null
                                 type: 'button', onClick: @onDraft,
                                     if @state.saving
-                                        span null,
-                                            img
-                                                src: 'images/spinner-white.svg'
-                                                className: 'button-spinner'
+                                        span null, Spinner(white: true)
```
frankrousseau: Shouldn't we put the span in the spinner widget?

client/app/components/menu_refresh_indicator.coffee
```diff
@@ -0,0 +1,68 @@
+{span, button, img} = React.DOM
+
+LayoutActionCreator  = require '../actions/layout_action_creator'
+
+{Spinner} = require './basic_components'
+
+module.exports = React.createClass
```
frankrousseau: I think a comment about what the component does could be interesting. What do you think about it?

client/app/components/compose.coffee
jsilvestre: I don't know, I just replaced what existed with the spinner component. Didn't rework the code around it. The span could simply be removed imo, but haven't digged into it.

frankrousseau:
Where is the reload button now?

jsilvestre:
I moved it at the right bottom. It is a very temporary location, @m4dz is going to make some room elsewhere for it. The idea was to move the button from a place where it was in a specific account context to a place where it's app-wide, since it refreshes *all* accounts.

frankrousseau:
I find it. I'm not sure putting it there is the right place. Maybe at the right of the account button would be a proper place. 

Here are other feedback don't worry much about them, I will add them when @poupotte will review the feature:

* Information display make the waiting experience really nicer.
* It says it check the Trash but not the Inbox (tested with our test account)
* The initalizing step is very long compared to other. Can't we split this step?

jsilvestre:
I can work a bit around the ImapReporter in order to split the initialization step, not sure how much work it requires. I will work on that next week. 
Btw, your comment in the PR has been taken care of.

frankrousseau:
Ok let's merge this and discuss improvements later.

frankrousseau:
Thank you @j

PR #380, opened 2015-04-24
Add email signature

frankrousseau:
Allow the user to save a signature at the account level. Once done, all composed message starts with a signature.

What is missing: position of the cursor is not proper when there is a signature.


client/app/components/compose.coffee
```diff
@@ -311,9 +312,13 @@ module.exports = Compose = React.createClass
...
         else
-            state = messageUtils.makeReplyMessage @props.selectedAccountLogin,
-                @props.inReplyTo, @props.action,
-                @props.settings.get('composeInHTML')
+            state = messageUtils.makeReplyMessage(
+                @props.selectedAccountLogin,
+                @props.inReplyTo,
+                @props.action,
+                @props.settings.get('composeInHTML'),
+                @props.selectedAccount.get('signature')
```
aenario: This serie of components get passed the selectedAccountID, all the accounts, the selectedAccountLogin and now the selectedAccount itself.

I am not sure which system is better, but we really ought to decide 

EIther we pass the `selectedAccountID` & `accounts` and then do `accounts.get(selectedAccountID).get('login')` and `accounts.get(selectedAccountID).get('conversation')`

Or we  pass the selectedAccount itself and get the ID from it.

Ping people working more on the frontend @clochix and @jsilvestre  ?

client/app/components/basic_components.coffee
```diff
@@ -51,9 +51,14 @@ Tabs = React.createClass
     render: ->
         ul className: "nav nav-tabs", role: "tablist",
             for tab in @props.tabs
+                if tab.class?.indexOf('active') > 0
```
aenario: should be >=  to work if active is the first class of the tab

client/app/components/compose.coffee
jsilvestre: That's typically the kind of things that will be fixed with the refactoring next week. It will depends on the component needs, I have not checked the whole component to decide what is the best option. I believe it needs to know all the accounts (to be able to switch), so the use of  `selectedAccountID` seems to be the best option.

frankrousseau: I think it's best to pass the selectedAccount object and retrieve fields from it.

client/app/components/basic_components.coffee
frankrousseau: Ok.

client/app/components/compose.coffee
frankrousseau: What refactoring are you talking about @jsilvestre ?

clochix: We already have the account list (for account selector) and current account id, so I think adding `selectedAccount` to the props is useless. You can use `@props.accounts.get(@props.selectedAccountID).get('signature)'`.

frankrousseau: Ok @clochix

frankrousseau:
I think it's good to merge now.

PR #384, opened 2015-04-27
Quickfixes for account config

clochix:
Fixes some errors and warning introduced by account config refactoring.


jsilvestre:
+1 I might need that as soon as possible ;-)

Issue #125, opened 2015-01-26
Display date of last mailbox refresh


poupotte:

poupotte:
It works well, thanks @jsilvestre .

PR #386, opened 2015-04-27
Added: last refresh date as tooltip on refresh button. Refs #125.

jsilvestre:
During review you will probably wonder why did I use `title` rather than a shiny tooltip. The reason is that tooltips with dynamic content are not possible right now with the way the application uses tooltips.

**Short answer**
We can't do tooltips with dynamic content right now due to the fact aria-tips changes the DOM and it messes React up. I can do something about it, but I feel like there are more important things to do.

**Long answer**
The tooltips library re-append all the element with `[role="tooltip"]` to `body` so it's relative to the body and not the parent's element it has been put in (thus, it's easier to position and animate). That means you break React's virtual dom, but that's okay when the tooltips never change.
When their content changes, it must be done the React way, with a source of truth you mutate. In our case, it would probably be a Flux store. That's a bit of work since I'm not a 100% sure of what to do.

Also fixed a bug in the initial state of the application (no account created yet).

### DIFF
--- a/client/app/locales/de.coffee
+++ b/client/app/locales/de.coffee
@@ -61,6 +61,7 @@ module.exports =
       "menu refresh initializing": "Initializing..."
       "menu refresh cleaning"   : "Cleaning..."
       "menu refresh indicator"  : "%{account}: %{mailbox} (%{progress}%)"
+      "menu last refresh"       : "Last refresh on %{date}."
 
       # List
       "list empty"              : "Keine E-Mail in diesem Postfach."
--- a/client/app/components/menu.coffee
+++ b/client/app/components/menu.coffee
@@ -106,10 +106,13 @@ module.exports = Menu = React.createClass
                         @getAccountRender account, key
                     .toJS()
 
-            RefreshIndicator
-                refreshes: @props.refreshes
-                mailboxes: @props.selectedAccount.get('mailboxes')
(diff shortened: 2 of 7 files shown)


client/app/locales/de.coffee
```diff
@@ -61,6 +61,7 @@ module.exports =
       "menu refresh initializing": "Initializing..."
       "menu refresh cleaning"   : "Cleaning..."
       "menu refresh indicator"  : "%{account}: %{mailbox} (%{progress}%)"
+      "menu last refresh"       : "Last refresh on %{date}."
```
frankrousseau: Do not add wrong translation.

jsilvestre: what do you mean?

frankrousseau: I mean that Last refresh is not german.

jsilvestre: If I don't put something, the user will see "menu last refresh", so I'd rather put the proper english string into the german localization.

frankrousseau:
We have to make some tests with Transifex. I'm afraid that with this behavior Transifex think that these keys are properly translated.

frankrousseau:
We'll see later for this problem of locale keys. It's more important that @poupotte can check the result of your dev.

PR #387, opened 2015-04-27
Add unit tests to client (do not merge)

frankrousseau:
First attempt to add unit tests on the client side. It works nice with the `jsdom` library. But brunch messes up with require, so we can't use it right now (require are runned even if they should not be executed).

That's something the Brunch team is working on. So at some point we can expect it will be solved.

Any ideas to fix this problem are welcome.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -1,0 +1,42 @@
+### Unit tests
+
+Running unit tests requires [io.js](https://iojs.org/en/index.html) and an
+additional parameter to setup `jsdom`.
+
+### Use io.js
+
+To use `io.js` it's recommended to run it via a node environment manager like
+[nvm](https://github.com/creationix/nvm).
+
+#### Install NVM (requires bash)
+
+```bash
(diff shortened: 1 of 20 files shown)


jsilvestre:
I can't wait we can use browserify with Brunch, that would make things much cleaner.

frankrousseau:
Good points @jsilvestre. There is a plugin called brunch-browserify. Do you know it?

jsilvestre:
Yes, but it has serious limitations. Browserify support should be native in the next major version of Brunch, I must take some times to try it out.

frankrousseau:
About jsdom the 3.x version requires more configuration and is not maintained anymore. So, I prefer to use the latest one.

jsilvestre:
That gives me a good reason to install io.js ;-) It's easy to do with https://github.com/tj/n

frankrousseau:
I close this one. We'll work on it when Brunch will be ready.

Issue #271, opened 2015-04-02
Infinite scroll is broken when displaying conversations


clochix:
In a mailbox, I have only one conversation, with 5 messages.
The “Load more messages” link is displayed, because it thinks we are only displaying one out five messages.

clochix:
I made a fix that sums up the number of messages inside each conversation.
It's better, but still false, because if some messages inside the conversation are in other malboxes, the sum will be wrong. We should also compute (server side ?) the number of messages of the conversation that are inside the current mailbox. /ping @aenario

aenario:
How about checking for the link : next in the answer instead of counting things ?

clochix:
Ok, I'll try to remove the code that count messages and see if it works. Otherwise, I'll come back to you.

clochix:
As a side effect of working on filtering, I found, at last, why using the link didn't work ☺

clochix:
@poupotte I updated the code used to determine if there are more messages to show, it should now be simpler and works in more cases. Crossing fingers.

poupotte:
It seems still broken :(

If I delete several messages/conversations, it displays 'This is the end of the road' instead of retrieve next messages.

poupotte:
Infinite scrool works for me. Thanks @clochix  and @aenario.

PR #391, opened 2015-04-28
Refactoring of message action - WIP

aenario:
So ....

This patch gives up on most REST principles and use action-based routes instead. The client is updated to follow the same logic and group all action-related code in message_action_creator. 

Server tests are broken, I am fixing them now. 

I'd recommend reviewing commit-by-commit.


frankrousseau:
There is another things than your refactoring like infinite scroll fix, locales, etc. Could you explain your changes?

client/app/components/basic_components.coffee
```diff
@@ -125,7 +125,41 @@ FormButtons = React.createClass
 
         div null,
             div className: 'col-sm-offset-4',
-                FormButton formButton for formButton in @props.buttons
+                for formButton, index in @props.buttons
+                    formButton.key = index
+                    FormButton formButton
+
+MenuItem = React.createClass
```
aenario: I made a few changes in the toolbar & toolbox components so that they always delegate the actual action to their parent component, because it depends on which view is showing them. However, i could not make them below 80 char, so I made the MenuItem & cie components to clean up these components. (https://github.com/cozy/cozy-emails/pull/391/files#diff-d3d0c90ae3619da533428714ee04cd31R44)

client/app/locales/de.coffee
```diff
@@ -231,12 +231,12 @@ module.exports =
       "message action sent ko"      : "Fehler beim Nachricht senden: "
       "message action draft ok"     : "Nachricht gespeichert"
       "message action draft ko"     : "Fehler beim Nachricht speichern: "
-      "message action delete ok"    : "Nachricht “%{subject}” gelöscht"
-      "message action delete ko"    : "Fehler bei Nachricht löschen: "
-      "message action move ok"      : "Nachricht “%{subject}” wurde verschoben"
-      "message action move ko"      : "Fehler beim verschieben von Nachricht “%{subject}”: "
-      "message action mark ok"      : "Nachricht markiert"
-      "message action mark ko"      : "Fehler beim Nachricht markieren: "
+      "message delete ok"    : "Nachricht “%{subject}” gelöscht"
```
aenario: MessageActionCreator now handles all action success / failure / undo, so its easier if they have a key of the same form
`type action ko/ok`

client/app/stores/message_store.coffee
```diff
@@ -147,14 +144,16 @@ class MessageStore extends Store
                     [key, value] = p.split '='
                     value = '-' if value is ''
                     _params[key] = value
-            else
+            else if messages.mailboxID
```
aenario: The infinite scroll breaks after an action, because we use RECEIVE_RAW_MESSAGES there. I thought it was due to my changes and made the fix. Then found out it was already there in #271

client/app/actions/message_action_creator.coffee
```diff
@@ -180,3 +40,290 @@ module.exports =
...
+                # refetch the batch from the server for update
+                MessageActionCreator.refresh target
+                LAC.alertError alertMsg
+            else
+                MessageActionCreator.receiveRawMessages updated
+                unless target.undeleting
+                    LAC.notify alertMsg,
+                        autoclose: true,
+                        actions: [
+                            label: t 'conversation undelete'
```
clochix: This allow to undo move, so maybe using something like "undo last action"

client/app/stores/message_store.coffee
```diff
@@ -375,6 +372,21 @@ class MessageStore extends Store
 
         return _conversationMemoize
 
+    getMixed: (target) ->
```
clochix: Could you add a comment, please ?

clochix:
Code seems good, lets merge and test ! Thanks @aenario.

PR #396, opened 2015-04-29
Fixes filters

clochix:
- include last day of date range ;
 - fix pagination when filtering on sender or recipient.


frankrousseau:
Thank you @clochix

Issue #288, opened 2015-04-07
Have a look at picture import inside the RTE


clochix:

clochix:
The first step is done, I investigated and know how to add an image to a message from an external source. It works for now with Web Activity to import an image from the File app, I need to work with @Benibur to use its popup that allows to select images from the photo application.

We should also handle the images dropped into the editor from the window manager of the user. I have some code for this, but it needs refactoring. Should I open another ticket, or do we keep this one ?

clochix:
I added this to current sprint, as I already did half of the work for the image drag and drop inside editor, and would like to finish it.

frankrousseau:
@clochix It's better if you fix current issues (realted to quick search). Demo day is thursday.

clochix:
@poupotte could you give it another try, I hope I fixed the bug you hit during demo.

poupotte:
It works well even if focus isn't on content mail. Thanks @clochix

PR #397, opened 2015-04-29
Allow to drop image inside Compose editor, refs #288

clochix:
When dropping a file inside compose editor, add it to the message attachments.
If file is an image and editor has focus, also insert the image into message body.


jsilvestre:
I reviewed this PR, it works, except that the limitation (the editor must have focus) is annoying. @clochix told me he was working on it.

frankrousseau:
@poupotte do you want it for this sprint?

poupotte:
@frankrousseau , yes,  in my mind, we can add it for the current sprint.

Issue #413, opened 2015-05-03
Deleting several emails should remove them instantly from the list


frankrousseau:
When I delete several emails, the deletion works but the experience is poor. It very long to remove every emails.

clochix:
How do you delete messages from the list ? When I select some messages and click on the trash in toolbar, they are immediately removed from the list (deleting message on the server takes more time, so the confirmation notification is displayed a few seconds later).

Deleting them with “delete conversation” in action menu seems broken.

frankrousseau:
I select something like 20 messages and click on the trash button.

poupotte:
Mmmm... I don't arrive to reproduce this issue. 
@frankrousseau , You have this behavior at each deletion or it's a random behavior ?

frankrousseau:
You should try on an online Cozy. Everything works well at the beginning but the messages come back then are deleted.

poupotte:
Ah ok, at the beginning you see your messages disappear but sometimes its reappear. Thanks for precision.

I can reproduce it. It's probably due to requests crossing.

clochix:
I'm still unable to reproduce. @frankrousseau, do you still have this problem with the last version of the application ?

frankrousseau:
Yeah the messages still come back when I delete too much messages in a row.

poupotte:
It should be fixed in the new version. @frankrousseau , can you test it on your Cozy ?

Issue #405, opened 2015-04-30
Wrong In-reply-to header


clochix:
When composing a new message, its `in-reply-to` header is set to `<>`, and all messages with this header are attached to the same conversation.

poupotte:
It seems work. Thanks @clochix

Issue #84, opened 2014-12-15
Allow to use a login distinct from the address


clochix:
Add a login field to IMAP and SMTP parameters

clochix:
I added Advanced SMTP parameters, that allow:
 - to select authentication method. It's useful when using a local SMTP server that don't require auth;
 - to use another login / password for SMTP server.

Dou you think I should do the same for IMAP ? Imho, the main login / password of the account should be enough to connect to IMAP server.

(the code will be on next merge)

frankrousseau:
I think we should do it when an user will ask it. We should not add too much features too early.

clochix:
@poupotte STMP advanced parameters have been added some times ago, and we won't add IMAP advanced parameters for now, so could we close this one ?

clochix:
I re-open this as one user rised this issue yesterday.

PR #418, opened 2015-05-04
Minor fixes

clochix:
This request should fix the following problems:
 - sometime, when dropping an image hover the compose editor, its content wasn't refreshed. I now force the refresh once the image has been loaded;
 - if there is a signature at the bottom of the message, insert image before it (previously, image was inserted on the bottom of the message, no matter if there where a signature);
 - fix signature separator ("^-- \n" instead of "^-- --\n");
 - fix the view used to filter message on dests: some messages (loaded from the fixtures for example) don't have a `cc` field, this broke the view.
 - restore IMAP advanced parameters. This was done for #84, but lost during account component refactoring
 - fix in-reply-to header, refs #405 
 - delete conversation from list menu didn't work, refs #413


clochix:
@aenario I resolved the conflict with the development branch, so you should be able to review now.

clochix:
I just added a commit that restore IMAP advanced parameters. This was done for #84, but lost during account component refactoring.

Issue #421, opened 2015-05-05
[regression] unable to get raw message


clochix:
The item in message menu doesn't do anything.

PR #423, opened 2015-05-06
Minor fixes and build

clochix:
- getting raw message source was broken, refs #421;
 - when deleting a message without subject, the notification contained “undefined” → fixed;
 - fix a syntax error on list selection (don't know how to reproduce, but the name of a variable was wrong);
 - minor improvement when displaying text-only messages;
 - build.


jsilvestre:
Looks good to me. Thank you!

Issue #424, opened 2015-05-06
Account autoconfig reset account form


clochix:
When editing an existing account, on blur in the address field, the account auto-discover is triggered. If no server is discovered, it delete current values.

poupotte:
It works well. Thanks @clochix

Issue #417, opened 2015-05-04
empty error message while setting up email


arthurlutz:
After getting errors on the imap port, I changed some settings, tried the check connection button again and got an empty red box and so fields highlighted. 

![screenshot from 2015-05-04 10 38 24](https://cloud.githubusercontent.com/assets/108437/7450824/cc0501e2-f249-11e4-97d8-3955ef610e0a.png)

clochix:
I'm unable to reproduce this one, so I updated the code at the only place I found where an empty notification may be thrown. I also prevented empty notifications to show up. Instead, the app will throw an exception that we can get in server logs to ease debug next time it happens.

arthurlutz:
Should cozy-email work with self signed certificates ?

clochix:
@arthurlutz I think that by default, Cozy will reject self-signed certificates.

The libraries we use allow to set advanced TLS options, so we should be able to allow the user to ask not to reject self signed certificates, or to register its certificate.

clochix:
In fact, I just saw in the code that we already allow self-signed certificates: https://github.com/cozy/cozy-emails/blob/master/server/imap/pool.coffee#L78
@aenario, could you confirm ?
IMHo, this is not secure, and we should add an option so user can choose to accept or reject untrusted certificates.

frankrousseau:
@aenario is in vacations. He will come back on monday.

Issue #232, opened 2015-03-25
Add 2 different fields for login and address


ZeHiro:
In version 0.6.9 of email app, there is only one field for login and adress. The problem is the following : 
- when filling adress with login@domain.tld, the connection to the imap/smtp server is not possible if the login is not the full adress
- when removing the @domain.tld, IMAP/SMTP servers are not filled automatically, and it is not possible to send an email, as there is no valid adress specified.

Proposal : 
- Add a second field "login"
- When address is filled in the form, login field is filled with the same value as address
- Login can be modified without impact on address

This should have inpact on couchDB structure (a new field adress could be necessary)

ZeHiro:
Hello,
do you have any plan to fix this?
I currenly cannot use the mail app to write mails (just reception).
Thanks

clochix:
Hello ZeHiro,
good news, this is in progress, see #84. Should be available in next version.

ZeHiro:
Cool. Thank you for the fast answer.
Good job on all the cozy project.

poupotte:
It's available on development branch. @ZeHiro, you can already test it if you installe emails in 'development' branch or wait the next version (at the end of next week).

ZeHiro:
@poupotte , I 've installed the dev branch, but it does not seem to work.
I tried to update the mail account (so I didn't create one), and the Check of the connection fails. 
It is like the imapLogin Field is not taken into account in the check of the connection, before enabling to save the new parameters.

Another bug I noticed : french locale for "Utilisateur IMAP (s'il est différent du login)" should be "Utilisateur IMAP (s'il est différent de l'adresse mail)"
The same for the SMTP login (not clear)

clochix:
My bad, checking the connection uses a different code path, where imapLogin wasn't used. I'll fix this and come back to you.
Could you try to save the account without checking the parameters, and tell me if it works ?

Thanks for your suggestions, I've updated the labels.

ZeHiro:
The imap logins and smtp logins were store in the DB, but I cannot check if this really works, as I cannot save the address : when saving the address, a connection check is done, and fails.

So I still don't have a proper mail saved in the "login" field of the DB, I cannot check : 
- if the right login is used to connect to imap
- the right login is used to connect to smtp
- I cannot test sending an email.

ZeHiro:
[...]

PR #429, opened 2015-05-07
First step of Web Intents

clochix:
This PR adds the code to use Web Intents developed by @Benibur to add a picture from Photo application into editor.

The code “pings” the Home to check if Web Intents are available.
I let some mock code used for test purpose, I'll remove it once we have everything we need in Home to test.

This PR also fixes minor bugs:
 - try to prevent empty error message on account creation, refs #417
 - use imapLogin when checking account parameters, refs #232
 - prevent reseting account parameters when editing account, refs #424
 
Have a nice review ☺


jsilvestre:
It looks good to me. Should I merge?

clochix:
Thanks ☺

Issue #431, opened 2015-05-07
[regression] Keyboard navigation is broken


poupotte:
I can't navigate in messages list with keyboard. 
(I don't receive error)

clochix:
Closing, it's a duplicate of #430 (same regression)

Issue #430, opened 2015-05-07
[regression] Delete message with keyboard is broken


poupotte:
Delete message with keyboard is broken. I haven't eror notifcation, but I receive this error in my console :
```
TypeError: messages[0] is undefined app.js:981:2
"TypeError: messages[0] is undefined" "https://bellot-zoe.cozycloud.cc/apps/emails/js/app.js" 981 2 TypeError: messages[0] is undefined
Stack trace:
_localDelete@https://bellot-zoe.cozycloud.cc/apps/emails/js/app.js:981:3
MessageActionCreator.delete@https://bellot-zoe.cozycloud.cc/apps/emails/js/app.js:740:16
module.exports.messageDeleteCurrent@https://bellot-zoe.cozycloud.cc/apps/emails/js/app.js:13194:1
root.mailkeys._binding.d.action@https://bellot-zoe.cozycloud.cc/apps/emails/js/vendor.js:52192:11
x@https://bellot-zoe.cozycloud.cc/apps/emails/js/vendor.js:52278:439
m.handleKey@https://bellot-zoe.cozycloud.cc/apps/emails/js/vendor.js:52284:1
y@https://bellot-zoe.cozycloud.cc/apps/emails/js/vendor.js:52279:144
 "_localDelete@https://bellot-zoe.cozycloud.cc/apps/emails/js/app.js:981:3
MessageActionCreator.delete@https://bellot-zoe.cozycloud.cc/apps/emails/js/app.js:740:16
module.exports.messageDeleteCurrent@https://bellot-zoe.cozycloud.cc/apps/emails/js/app.js:13194:1
root.mailkeys._binding.d.action@https://bellot-zoe.cozycloud.cc/apps/emails/js/vendor.js:52192:11
x@https://bellot-zoe.cozycloud.cc/apps/emails/js/vendor.js:52278:439
m.handleKey@https://bellot-zoe.cozycloud.cc/apps/emails/js/vendor.js:52284:1
y@https://bellot-zoe.cozycloud.cc/apps/emails/js/vendor.js:52279:144
"
```

poupotte:
@clochix , Do you think the lack of notification may be related with #417 ?

clochix:
I'm on it, just a call forgotten during mass deletion refactoring.
No notification because this is an error in our code that should never happen, so it's not catched. Fix ASAP.

clochix:
[...]

Issue #404, opened 2015-04-30
Display first unread message in conversation


clochix:
When displaying a conversation with several unread messages, we should open the oldest unread message instead of the newest message.

poupotte:
It works well for me. Thanks @clochix .

Issue #403, opened 2015-04-30
Don't mark all conversation as read when opening a message


clochix:
When I open a conversation with several unread messages, all messages are marked as read. We should only mark a message as read when the user really opens it

poupotte:
It works for me. Thanks @clochix

PR #432, opened 2015-05-07
Improve display of conversations

clochix:
- open the first unread message of conversation, refs #404 ;
 - don't mark as read all messages of conversation when opening it, refs #403 ;
 - fix regression introduced in previous build, that cause #430 and #431;
 - draft auto-save broke embedding of images. Should be fixed ;
 - previous fix of `in-reply-to` sometime made send fail. Should be fixed ;
 - try to prevent flashing of Seen flag, refs #274 ;


jsilvestre:
Thanks!

PR #445, opened 2015-05-13
Bugs: fix compose btn

m4dz:
* fix compose btn action
* improve the panel mode
* fix the embed mode width


frankrousseau:
Thx @m4dz

PR #449, opened 2015-05-15
Changed: refresh button doesn't do anything (for now).

jsilvestre:
@Benibur suggested we don't show at all the imap refresh to the user, because it's hard to it right (if you want details on why, call me it's be quicker). Instead, we refresh the client (browser)'s cache with server's data.

Since it's already done throught realtime, I made the button...do nothing. If that works better, we could improve in that direction later on, but today is release day, so this is more a quickfix than anything else.


clochix:
Looks good to me, that's waht we discussed on IRC. Thanks @jsilvestre ☺

frankrousseau:
Not so good, I can't read my emails anymore.

frankrousseau:
(on my second account).

clochix:
We need to merge quickly, to test and find bugs. My latest PR contains bugfixes for problems I got after the previous merges.

PR #451, opened 2015-05-15
Fix flags management and minor errors

clochix:
Regarding flags management, I did two updates : 
 - update flags in all boxes the message belongs
 - update keywords before flags, because on some messages, every time we updated the flags, there was a subsequent IMAP instruction to add the OLD keyword, and this keyword removed all other flags. @aenario, it would be great if you could have a look at this when you come back.

Rest of the PR is mostly fixes of client code syntax errors and regressions.


frankrousseau:
Thank you @clochix for the many fixes.

PR #462, opened 2015-05-18
Fix display of next message on deletion

clochix:
This PR also merge master branch from Cozy, so it contains some commit from master that where not on development branch.


frankrousseau:
Thank you @clochix

PR #469, opened 2015-05-20
Fix en locale export

m4dz:


aenario:
Was merged directly into master

PR #471, opened 2015-05-20
Lets users formatting more consistent

m4dz:
Message header block and message details popup now uses the exact same
way to render participants. This will be simpler to upgrade the
component consistently.


m4dz:
Oups, branch deleting too fast :stuck_out_tongue:

PR #472, opened 2015-05-21
Fixes and code cleanup

clochix:
Fixes some issues (see detail in commit messages) and cleanup code of initialize.coffee


clochix:
When it will be good for you, I'll add a build.

clochix:
I only delete without asking when creating a new draft. When editing a draft, application will always ask confirmation before deleting it.

PR #480, opened 2015-05-26
Faster refreshes for servers supporting RFC4551 (Gmail / OVH / ...)

aenario:
@clochix , I am not sure if the changes to `client/app/actions/layout_action_creator.coffee` are 
A - a good idea (refresh the box everytime it is re-opened
B - the Right-Place-To-Do-It with react / flux


frankrousseau:
I noticed several cosmetic problems. Everything look fine but I don't know the code well enough to review it properly. So, I will merge it once your cosmetic fixes done.

clochix:
I don't know if refreshing every time the box is opened is a good idea, but I think that if we want to do so, it's not the right place to do so. Your code will only be called if current query is already in local cache, which should almost never happen.

We should try to find a way to trigger the refresh when the selected Mailbox change.

aenario:
There is a bug preventing import of one account. Please dont merge right now

PR #484, opened 2015-05-27
Improve dialog window and minot fixes

clochix:
This is mostly a refactoring to use a popin dialog window instead of `window.confirm`. For this, I add to add the display of the window inside `application.coffee`, so I updated all other calls of the modal to use the same code.


frankrousseau:
Looks great. Can you split the compose file in smaller ones?

clochix:
I'm refactoring Compose component to split it into smaller files. Can I suggest to merge this one, so our beloved POs can test ? I'll create another PR for the refactoring, so we have enough time to check it doesn't introduce regressions.

frankrousseau:
Ok.

frankrousseau:
@m4dz @aenario can you review it quickly? If it's not the case, we can merge it directly.

aenario:
Good for me. Merging now.

PR #489, opened 2015-05-29
Faster page load : Dont include contacts pictures

aenario:


frankrousseau:
Thank you for the change.

PR #502, opened 2015-06-04
Message Action Fixes

aenario:
Replace the message-by-message logic with mailbox-by-mailbox and
messages batching. Also correct some logic error with handling of
message in weird states (trash + other box).


aenario:
Testing on my cozy right now and I still have some bugs. Please dont merge.

aenario:
Actually, the bugs are "only" display and realtime issues and will be solved during this sprint. Good to merge & publish as-is. @poupotte , could you install & review that version ?

poupotte:
Actions as deletion or mark as read work well for me. However, filters don't work well. Indeed, messages displayed after a selection or an unselection don't always correspond to the filter but I'm not sure if this issue is due to this PR.

frankrousseau:
ok let's merge it.

PR #503, opened 2015-06-05
Refactoring of `application.coffee` and tests fixes

clochix:
I split the big `application.coffee` so it's only in charge of layout now, and created a `panel.coffee`. This probably introduced a lot of regressions, so need to be well tested before being released.

I also updated the client tests, but they are not ready to be re-activated in Travis, as they keep failing randomly sometime.


frankrousseau:
Thank you @clochix. Say welcome to one week of debug!

PR #531, opened 2015-06-29
Improve error handling (free use case)

aenario:


client/app/stores/account_store.coffee
```diff
@@ -141,6 +141,7 @@ class AccountStore extends Store
 
         handle ActionTypes.NEW_ACCOUNT_ERROR, (error) ->
             _newAccountWaiting = false
+            error.uniq = Math.random()
```
jsilvestre: Where do you use .uniq afterwards?

aenario: nowhere, it just force Panel.shouldComponentUpdate to re-render

jsilvestre: Okay, I see. Could you please add a comment about it? I feel like it's not trivial.

aenario:
done

aenario:
added some travis fixes.

PR #558, opened 2015-07-14
Correct error message

robinmoussu:
Hi,
I'm quite shure the error message displayed when mail need to be refresh is not in a good English.
Since that patch is realy trivial, you can do whatever you want with it, even change my name to yours ;)

### DIFF
--- a/build/server/views/reindexing.jade
+++ b/build/server/views/reindexing.jade
@@ -38,4 +38,4 @@ html
     div.reindexing-message
         img.spinner(src="images/spinner.svg")
         h1 We need to reindex your emails.
-        p This page with refresh in a minute.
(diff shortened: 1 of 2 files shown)


clochix:
Thanks for the help @robinmoussu ! Our technical experts will review it and merge it if it fits our high level of quality expectations ;-)
BTW, @aenario, why is this message hard coded in the template, and not in an l10n file ? Just asking :-P

aenario:
Hi @robinmoussu , this was merged into development and will be released at the beginning of next week.

@clochix , because at that point we dont have the cozy_instance so we dont know which locale to use (although, we should probably use the browser one).

Thanks !

PR #565, opened 2015-07-17
Various fixes

frankrousseau:
Hello,

Here are various fixes I did:

* Update medium editor version (it fixes glitches and unconsistent styles)
* Fix mail stylesheet to properly render mailchimp emails
* Simplify add/show contact ux in message headers (remove tooltip and full link and add two icons for add and show actions)


clochix:
@frankrousseau I remember we added a little patch to medium code to prevent inserting paragraphs when users hit enter. I don't know if this hack is still needed, or if you solved your display concerns with some CSS in the message.

PR #577, opened 2015-07-22
Misc fixes

clochix:


client/vendor/plugins/keyboard/init.js
```diff
@@ -83,7 +83,7 @@ if (typeof window.plugins !== "object") {
       'enter': {
         name: "Display current message",
         action: function (e) {
-          var btnConfirm = document.querySelector('.modal .modal-footer [class="btn btn-cozy"]');
+          var btnConfirm = document.querySelector('.modal .modal-footer .btn.btn-cozy');
```
aenario: How about giving this button an ID / unique class ?

client/app/components/account_config_main.coffee
```diff
@@ -1,3 +1,4 @@
+AccountActionCreator = require '../actions/account_action_creator'
```
aenario: We both did this fix, so the import is now twice in the file (here and at line 10)

frankrousseau:
Thank you @clochix. We have to ship tomorrow. That's why I merge this PR.

PR #579, opened 2015-07-23
Improve reply experience

frankrousseau:
Currently answering to email with Cozy is hard. The compose form doesn't perform well inside the conversation. The fact that everyone "reply on top" makes the "reply on bottom" by default annoying. The Medium editor was buggy and the message was not well rendered in the blockquote. So here are the changes I made:

* All composing actions use the compose form (including reply, reply all and forward). 
* They are all triggered via a route changing (browser tabs can be used).
* Reply on top is the default.
* Styles in composing are less restrictive.
* Medium editor is up to date.
* Non related extra: I fixed the contact button tooltips.

The new medium editor puts back the wrong behavior of creating a paragraph each time `enter` is pressed. I didn't manage to fix it. I'm speaking with the medium editor authors to find a right solution: https://github.com/yabwe/medium-editor/issues/70#issuecomment-124001964


client/app/stores/account_store.coffee
```diff
@@ -268,15 +287,34 @@ class AccountStore extends Store
...
     isMailboxRefreshing: (mailboxID)->
         _mailboxRefreshing[mailboxID] > 0
 
+
     getMailboxRefresh: (mailboxID) ->
         if _mailboxRefreshing[mailboxID] > 0 then 0.9 else 0
 
+    # Returns corresponding mailbox for given message and account.
+    getMailbox: (message, account) ->
+        for boxID of message.mailboxIds when boxID in account.favorites
```
aenario: either message is a Immutable.Map and you need to use .get(mailboxIds) or it's not and you need to use direct call. You mix both in this function.

client/app/components/compose.coffee
```diff
@@ -162,24 +171,62 @@ module.exports = Compose = React.createClass
...
                 div className: 'clearfix', null
 
-    onCancel: (e) ->
-        e.preventDefault()
 
-        # Action after cancelation : call @props.onCancel
-        # or navigate to message list
-        if @props.onCancel?
-            @props.onCancel()
+    # Build message hash with the many ids required: account ID, mailbox ID,
```
aenario: This whole block probably doesn't belong in a view component, the best way would be for the LayoutStore to generate and keep this "back url", and the panel component to pass it as a prop to the compose component.

frankrousseau:
Thank you for the comments. I will fix it soon.

PR #580, opened 2015-07-23
Fix account update

clochix:
Fix a regression (non-oauth account edit was broken) and a long standing bug on update of special folders.


frankrousseau:
Thank you @clochix

PR #586, opened 2015-08-06
Refactoring : with Processes and a Store

aenario:


frankrousseau:
Thank you @aenario! Let's go for testing now

PR #607, opened 2015-08-27
Remove patchConversation

aenario:
By fetching messages in the proper order (oldest first), we avoid the need for patch conversation (except after first import).

Also included : dont use the proxy CDN fonts to prevent flash of fonts (ref : https://github.com/cozy/cozy-proxy/pull/166)


frankrousseau:
Thank you  @aenario.

PR #633, opened 2015-09-14
Dev/autocompletion

poupotte:


client/app/components/compose.coffee
```diff
@@ -52,6 +52,10 @@ module.exports = Compose = React.createClass
 
 
     shouldComponentUpdate: (nextProps, nextState) ->
+        if @state.saving isnt nextState.saving
```
aenario: why do you need to add this ?

poupotte: The problem is : 
 * Application start automatic saving
 * You add a contact  : @state is updated
 * Application finish automatic saving
 * New contact isn't in nextState , thus contact just added disappears.

aenario: I think, you should probably do it there : 
https://github.com/cozy/cozy-emails/blob/f490f6139aa177d490be8ac98f767ec7d4a8df4b/client/app/components/compose.coffee#L459

Issue #634, opened 2015-09-14
Improve mailbox menu


poupotte:
If you have two mailbox with the same first letter, it's difficult to distinguish them in left menu.

Issue #627, opened 2015-09-10
Improve attachments icon visibility


poupotte:
It should be nice to improve attachment visibility, espacially on conversation.

kelukelu:
To improve the visibility in a conversation, I suggest to show the attachment icon event when the conversation are collapsed.
<img width="651" alt="screen shot 2015-09-15 at 10 58 59" src="https://cloud.githubusercontent.com/assets/7029620/9872755/b711b2a6-5b9b-11e5-87de-324679e36583.png">
If it is not visible enough, adding a lightgrey background color can be a good idea

PR #637, opened 2015-09-16
Development

poupotte:
* Fix search
* Display attachment icon even if message is collapsed, refs #627 
* Put the active account at the top of left menu, refs #634


client/app/components/menu.coffee
```diff
@@ -133,6 +133,13 @@ module.exports = Menu = React.createClass
                 if @state.accounts.length
                     @state.accounts.map (account, key) =>
                         @getAccountRender account, key
+                    .sort (account1, account2) =>
+                        if @props.selectedAccount?.get('id') is account1._store.props.key
+                            return -1
+                        else if @props.selectedAccount?.get('id') is account2._store.props.key
+                            return 1
+                        else
+                            return 0
```
aenario: might as well sort by ID or name otherwise to have a stable order.

client/app/components/message-list.coffee
```diff
@@ -58,6 +58,10 @@ module.exports = MessageList = React.createClass
                 @setState allSelected: false, edited: false
 
     render: ->
+        href = window.location.href
```
aenario: Could you move this into the router ? I feel it doesnt belong here.

Issue #623, opened 2015-09-08
Order of mailboxes in the list is wrong so the tree appears wrong too.


aenario:

frankrousseau:
I changed that. It now displays mailboxes in the alphabetical order.

Issue #594, opened 2015-08-11
Error creating folder


Peltoche:
Salut,

Le message d'erreur:
```
Error creating folder Argh, I'm unable to perform this action, please try again
```

La procedure:

- Creer un nouveau dossier dans choisir de parents
- Creer un autre dossier avec en parents le dossier precedent.

poupotte:
This issue should be now fixed. @peltoche, is it fixed for you ?

poupotte:
@aenario, this issue seems still available. Could you check it ?

clochix:
Hi @Peltoche,
This should be fixed for some time now, is it ok for you?

Peltoche:
Yes, all good! Thanks!

PR #638, opened 2015-09-17
Various mailbox fixes

aenario:
Error with server using "." as mailbox path delimiter ref #594
Error with order of mailboxes ref #623
Error when a box was destroyed via IMAP, it was kept in favorites
Error when a box was destroyed just after message were moved to it, the messages were not removed.


frankrousseau:
Thank you @aenario !

PR #672, opened 2015-10-30
use starred instead of stared, refs #665

poupotte:


aenario:
Did you make the change in transifex too ?

PR #693, opened 2015-12-02
SMTPConnection dep fix w/ npm3

m4dz:


nono:
The build is red, because tests on travis still run for JS files, not coffee.

PR #715, opened 2016-01-07
node 4

nono:
- Update dependencies for node 4
- Fix an error for node 0.12 / 4 / 5 when using buffers for attachments


server/models/message.coffee
```diff
@@ -555,11 +556,17 @@ module.exports = class Message extends cozydb.CozyModel
         async.eachSeries attachments, (att, cb) =>
             # WEIRDFIX#1 - some attachments name are broken
             # WEIRDFIX#2 - some attachments have no buffer
+            # WEIRDFIX#3 - use a stream instead of a buffer
             # att.name = att.name.replace "\ufffd", ""
-            # attachBinary need a path attributes
             att.buffer ?= new Buffer 0
-            att.buffer.path = encodeURI att.name
-            @attachBinary att.buffer, name: att.name, cb
+            stream = simplebufferstream att.buffer
```
aenario: How about we move this part to cozydb ?

nono: @aenario if you think it make more sense to do that in cozydb, I can move this hack to cozydb.

nono:
I've moved the work-around for using buffers as attachments to cozy-db. We have to wait for a new release of cozy-db to check if travis build is green and merge this PR if so.

nono:
Cozy-db 0.1.9 has been published and the build is green on travis \o/

frankrousseau:
Thank you!

PR #727, opened 2016-01-19
Fix Gmail security

aenario:
The imap server returned by autoconfig has changed, so we lost the security warning for gmail.

Also added some explanations about the rare(r) case, where the user need to display the gmail unlock captcha even if they have activated access for less secure app.


client/app/locales/en.json
```diff
@@ -333,6 +333,8 @@
   "gmail security tile": "About Gmail security",
   "gmail security body": "Gmail considers connection using username and password not safe.\nPlease click on the following link, make sure\nyou are connected with your %{login} account and enable access for\nless secure apps.",
   "gmail security link": "Enable access for less secure apps.",
+  "gmail security body 2": "If you still cant connect, you might need to use the unlock captcha from the following link.",
```
frankrousseau: cant should be can't

PR #732, opened 2016-01-28
Several fixes

aenario:


client/app/components/toolbox_actions.coffee
```diff
@@ -1,166 +1,144 @@
...
-        items = [
-            MenuHeader key: 'header-mark', t 'mail action mark'
-
-            if not @props.isSeen? or not @props.isSeen
+        message = @props.mode is 'message'
+        conversation = @props.mode is 'conversation'
+        Menu
+            icon: 'fa-cog'
+            direction: if @props.direction is 'right' then 'right' else 'left'
+        ,
```
nono: @aenario I'm a bit lost here. What means the comma on this line?

client/app/components/toolbox_actions.coffee
```diff
@@ -1,166 +1,144 @@
...
+
+            if message
+                MenuHeader key: 'header-mark', t 'mail action mark'
+
+            if message and @props.isSeen isnt true
```
nono: There is a good reason to write `isnt true` instead of `is false`? I'd have expected `isSeen` to be a boolean, but maybe it can take other values.

nono:
The code style looks good to me

client/app/components/toolbox_actions.coffee
aenario: React components takes an attributes object as first param and then the children as nullable varargs, I needed the coma to make coffee compile this part.

client/app/components/toolbox_actions.coffee
aenario: isSeen can be either true (all messages in conversation are seen), false (all messages are not seen), null (some are, some are not). I agree this is unclear and will add more comments.

PR #737, opened 2016-02-05
Fix rt import

aenario:


server/utils/constants.coffee
```diff
@@ -4,7 +4,7 @@ exports.MSGBYPAGE = 30
 # loads 200 ids in memory at once
 exports.LIMIT_DESTROY = 200
 # loads 30 messages in memory at once
-exports.LIMIT_UPDATE = 30
+exports.LIMIT_UPDATE = 50
```
nono: Comments don't say the same thing than code. You should remove them or make more generic (30 vs 50 above and 5 vs 1 just below).

PR #742, opened 2016-02-10
Fix download link on empty message

misstick:
Fix download feature : 
 - [x] link is broken
 - [x] cache local file url

### DIFF
--- a/client/app/components/file_picker.coffee
+++ b/client/app/components/file_picker.coffee
@@ -226,10 +226,8 @@ FileItem = React.createClass
         li className: "file-item", key: @props.key,
             i className: "mime #{type} fa #{iconClass}"
             a
-                className: 'file-name',
-                target: '_blank',
-                onClick: @doDisplay,
-                href: file.url
+                'className': 'file-name'
+                'onClick': @doDisplay
                 'data-file-url': file.url
                 'data-file-name': file.generatedFileName
                 'data-file-type': file.contentType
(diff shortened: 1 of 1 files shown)


client/app/components/file_picker.coffee
```diff
@@ -225,27 +227,26 @@ FileItem = React.createClass
...
 
             span className: 'file-size',
                 "\(#{(file.length / 1000).toFixed(1)}Ko\)"
             span className: 'file-actions',
                 a
-                    className: "fa fa-download"
-                    href: "#{file.url}?download=1"
+                    'className' : 'fa fa-download'
+                    'download'  : file.generatedFileName
+                    'href'      : getFileURL file
```
aenario: use `@state.tmpFileURL` here too

misstick: @aenario : it must be the real `file.url`here otherwise feature "download" doesn't work. See my rollback commit: https://github.com/misstick/cozy-emails/commit/b6378cf11ea2ed38160268dbac9e50a280ad5c10#diff-4ef6be8edecdebfd711979b18edc3434L232

aenario: If we use the download attributes, I think we dont need the ?download=1 anymore.

misstick: @aenario : you're right : it is exactly what this code do :)

aenario: Oh. Brain-bug,  sorry about that :-)

PR #744, opened 2016-02-12
Remove 2nd panel when global scope change

misstick:
Remove 2nd panel (conversation from an email selected into emailList) when : 
 - [x] the selected email is deleted,
 - [x] the mailbox filter is changed,
 - [x] the user change the mailbox
 - [x] the user delete a draft from message.editView


misstick:
Oups : message are no more saved in draft

PR #747, opened 2016-02-16
Autosave create un-useless draft 

misstick:
## Creating a new message
When I click on 'compose' : 
 - an empty draft is displayed

When I change the fields : to, cc, ccc, subject, message, attachment,
 - [x] message is saved on client,
 - [x] message is saved on server.
 - [x] a draft is created on 1st save,
 - [x] DO not create empty message

## Updating message in Drafts
I choose 'Draft' MailBox, and I click on a message in the list
 - [x] ComposeView is displayed with the right data

When I change the fields : to, cc, ccc, subject, message, attachment,
 - [x] message is saved on client,
 - [x] message is saved on server.
 - [x] a draft is created on 1st save,
 - [x] that draft is updated on other save

## Create message : Display ModalBox
 I save data when the component is unmounted 
 - [x] I show Modal when there are differences between client and server data
 - [x] nothing happen if I saved draft before component is unmounted

## Edit draft message : Display ModalBox
 I save data when the component is unmounted 
 - [x] I show Modal when there are differences between client and server data
 - [x] nothing happen if I saved draft before component is unmounted

When I click on a draft into the list and click on another draft :
 - [x] `subject` is updated
 - [x] `cc` is updated
 - [x] `bcc` is updated
 - [x] `content` is updated

Then I click on `cc` (field `cc` is visible) and change the draft selection :
 - [x] `cc` is hidden

Then I click on `bcc` (field `cc` is visible) and change the draft selection :
 - [x] `bcc` is hidden

## Basic message feature
 - [x] toggle between `cc`and `bcc` fields,
 - [x] attachements : add, remove, save,
 - [x] HTML text is style HTML,
 - [x] picture is still a picture

### DIFF
--- a/client/app/components/compose.coffee
+++ b/client/app/components/compose.coffee
@@ -132,7 +132,7 @@ module.exports = Compose = React.createClass
                 div className: 'compose-content',
                     ComposeEditor
                         id                : 'compose-editor'
-                        messageID         : @props.message?.get 'id'
+                        messageID         : @state.id
                         html              : @linkState('html')
                         text              : @linkState('text')
                         accounts          : @props.accounts
@@ -217,11 +217,16 @@ module.exports = Compose = React.createClass
         else if @props.inReplyTo?
             document.getElementById('compose-editor')?.focus()
 
+    isNew: ->
+        not @state.originalConversationID
+
+    unsetNew: ->
+        @state.originalConversationID = @state.conversationID
+
     componentDidUpdate: ->
         # Initialize @state
         # with values from server
-        if @props.settings.get('autosaveDraft') and not @state.id
-            @saveDraft()
+        @saveDraft() if @isNew()
 
         # focus
         switch @state.focus
@@ -256,7 +261,24 @@ module.exports = Compose = React.createClass
         LayoutActionCreator.displayModal params
 
     componentWillUnmount: ->
-        return if @props.lastUpdate is @state.date
+        success = (error, message) =>
+            msg = "#{t "message action draft ok"}"
+            LayoutActionCreator.notify msg, autoclose: true
+
+            # reload conversation to update its length
+            if message and message.conversationID?
+                cid = message.conversationID
+                MessageActionCreator.fetchConversation cid
+
+
+        if @props.lastUpdate is @state.date
+            message = @state
+            error = null
+            setTimeout ->
+                success error, message
+            , 0
+            return
+
 
         doSave = =>
             LayoutActionCreator.hideModal()
@@ -264,15 +286,10 @@ module.exports = Compose = React.createClass
                 if error? or not message?
                     msg = "#{t "message action draft ko"} #{error}"
                     LayoutActionCreator.alertError msg
+                    success error, message
                     return
 
-                msg = "#{t "message action draft ok"}"
-                LayoutActionCreator.notify msg, autoclose: true
-
-                # reload conversation to update its length
-                if message.conversationID?
-                    cid = message.conversationID
-                    MessageActionCreator.fetchConversation cid
+                success error, message
 
         init = =>
             @showModal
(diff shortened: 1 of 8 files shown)


client/app/components/compose.coffee
```diff
@@ -50,21 +50,129 @@ module.exports = Compose = React.createClass
...
+    isNew: ->
+        not @state.conversationID
 
     shouldComponentUpdate: (nextProps, nextState) ->
-        return not(_.isEqual(nextState, @state)) or
-            not (_.isEqual(nextProps, @props))
+        !!nextProps.accounts
 
+    componentWillReceiveProps: (nextProps) ->
+        if nextProps.message
```
aenario: You can probably avoid having to manage this situation (the compose component is injected with another message) by using smart key (compose-#{message.id} ?) attribute in panel.coffee

client/app/components/compose.coffee
```diff
@@ -50,21 +50,129 @@ module.exports = Compose = React.createClass
...
+                nextState.html = MessageUtils.wrapReplyHtml nextState.html
+
+    componentDidMount: ->
+        # scroll compose window into view
+        @getDOMNode().scrollIntoView()
+
+        # Focus
+        if not Array.isArray(@state.to) or @state.to.length is 0
+            setTimeout ->
+                document.getElementById('compose-to')?.focus()
```
aenario: Using autofocus would be cleaner. Can be done in another PR.

client/app/components/mails_input.coffee
```diff
@@ -25,10 +25,12 @@ module.exports = MailsInput = React.createClass
         state.unknown  = ''
         state.selected = 0
         state.open     = false
+        state.focus    = !!state.known.length
         return state
 
     componentWillReceiveProps: (nextProps) ->
-        @setState known: nextProps.valueLink.value
+        value = nextProps.valueLink.value
+        @setState known: value, focus: !!value.length
```
aenario: Use value.length isnt 0 instead of !!

client/app/utils/message_utils.coffee
```diff
@@ -175,11 +185,12 @@ module.exports = MessageUtils =
             when null
                 @setMessageAsDefault options
 
-        # remove my address from dests
-        notMe = (dest) -> return dest.address isnt myAddress
-        message.to = message.to.filter notMe
-        message.cc = message.cc.filter notMe
-        return message
+        # # remove my address from dests
```
aenario: Dont leave commented code, if you really really think we will need it in the future, leave a comment with a git commit number.

client/app/components/mails_input.coffee
```diff
@@ -127,11 +128,12 @@ module.exports = MailsInput = React.createClass
...
+        # Show field if results
+        className += ' shown' if @state.focus
+
         # don't display placeholder if there are dests
-        if knownContacts.length > 0
-            placeholder = ''
-        else
-            placeholder = @props.placeholder
+        hasNoContact = _.isEmpty knownContacts
+        placeholder = unless hasNoContact then '' else @props.placeholder
```
aenario: `if ... then ... else` is more readable than `unless ... then ... else`

client/app/components/mails_input.coffee
```diff
@@ -127,11 +128,12 @@ module.exports = MailsInput = React.createClass
             else
                 event.dataTransfer.dropEffect = 'none'
 
+        # Show field if results
+        className += ' shown' if @state.focus
```
aenario: Beware, focus and shown are two different things, this could cause confusion in the future.

client/app/components/compose_toolbox.coffee
```diff
@@ -38,10 +56,10 @@ module.exports = ComposeToolbox = React.createClass
                         button
                             className: 'btn btn-cozy-non-default btn-delete',
                             type: 'button',
-                            onClick: @props.onDelete,
+                            onClick: @delete,
                                 span className: 'fa fa-trash-o'
                                 span null, t 'compose action delete'
                     button
-                        onClick: @props.onCancel
+                        onClick: @props.cancel
```
aenario: Shouldn't it be @cancel

client/app/components/compose.coffee
misstick: @aenario : that's what I thought first, and I deleted this part of code; but il fact it's about editing 'compose', in Draft Mailbox. When the route change from 'draftMessageList' to 'editDraftIntoMessageList' the component isnt destroyed bu t refresh, so the data is initialized into `getInitialState` ans updated in `shouldComponentUpdate`.

client/app/components/compose.coffee
misstick: @aenario :+1: I didnt have enough time

client/app/components/compose_toolbox.coffee
```diff
@@ -5,31 +5,49 @@
...
+
+    delete: ->
+        @triggerAction 'delete'
 
+    cancel: ->
+        @triggerAction 'cancel'
+
+    triggerAction: (type) ->
+        @setState action : type
+        if (callback = @props[type]) and _.isFunction callback
```
aenario: Better to use `listener` than `callback` here. You can also use `@props[type]?()`. 
I feel `triggerAction` might be too much factorisation, as I often search for `props.xxxxx` in a component to see how its used, may be it would be cleaner to just copy paste 
```
@setState action: 'xxxx'
@props.xxxxx?()
``` 
in the functions above ?

client/app/components/compose.coffee
```diff
@@ -210,378 +323,120 @@ module.exports = Compose = React.createClass
...
-
-    onToggleBcc: (e) ->
-        toggle = (e) -> e.classList.toggle 'shown'
-        toggle e for e in @getDOMNode().querySelectorAll '.compose-bcc'
-        focus = if not @state.bccShown then 'bcc' else ''
-        @setState bccShown: not @state.bccShown, focus: focus
+    toggleField: (event) ->
+        ref = event.currentTarget.getAttribute 'data-ref'
+        view = @refs[ref]
+        value = !!!view.state.focus
```
aenario: use a single not instead of !!!

client/app/components/mails_input.coffee
misstick: @aenario : dont you prefer `not _.isEmpty` instead?

client/app/components/mails_input.coffee
misstick: @aenario : Note that it is `MailInput.focus` not a DOM event `input.focus()`

client/app/components/compose_toolbox.coffee
misstick: @aenario : in fact, I didnt want to repeat all the logic each time

PR #759, opened 2016-02-25
Loosing focus in message list when remove a message (Suppr)

misstick:
Actually `_fixCurrentMessage ` should do that stuff but it doesnt work.

When a message is deleted : 
 - [x] next message in the list should be focus,
 - [x] URL should be updated.

When a user navigate with keyboard into messageList : 
 - [x] next message should be focus when I press on `top` key,
 - [x] previous message should be focus when I press on `bottom` key.

### DIFF
--- a/client/app/utils/api_utils.coffee
+++ b/client/app/utils/api_utils.coffee
@@ -40,9 +40,7 @@ module.exports =
 
 
     getCurrentConversation: ->
-        conversationID = MessageStore.getCurrentConversationID()
-        if conversationID?
-            return MessageStore.getConversation(conversationID)?.toJS()
+        MessageStore.getCurrentConversation()?.toJS()
 
 
     getCurrentActions: ->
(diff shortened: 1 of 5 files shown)


frankrousseau:
Could you add comments telling what your function do and why you chose do it that way (if needed).

misstick:
@frankrousseau : I thought function names are the most explicit here, didnt it?

frankrousseau:
Yep, but you can explain how the function works and why you choosed to do it that way.

client/app/utils/api_utils.coffee
```diff
@@ -95,21 +93,21 @@ module.exports =
...
     messageNavigate: (direction, inConv) ->
-        if not onMessageList()
-            return
+        return unless onMessageList()
         conv = inConv and SettingsStore.get('displayConversation') and
             SettingsStore.get('displayPreview')
+
         if direction is 'prev'
-            next = MessageStore.getPreviousMessage conv
+            next = MessageStore.getNextConversation()
```
aenario: this seems reverted.

frankrousseau:
It's ok for me. @aenario Do you think we can merge it?

misstick:
@frankrousseau : I'm adding comments on store, to be more explicit as you asked before

misstick:
@frankrousseau, @aenario : it's OK for me

PR #760, opened 2016-02-26
FIX: Checkbox were not checked when selected

misstick:


client/app/components/message-list.coffee
```diff
@@ -84,8 +84,9 @@ module.exports = MessageList = React.createClass
             displayConversations : displayConvs
 
     # for SelectionManagerMixin
-    getSelectables: (props = @props, state = @state) ->
-        state.messages.keySeq()
+    getSelectables: ->
+        @state.messages.map (message) ->
+            message.get('id'): message
```
aenario: This gives me a syntax error ?

PR #766, opened 2016-02-29
[chore] webpack migration

m4dz:
This PR introduces deep changes:
- use webpack instead of Brunch for frontend compilation
- migrate deps to NPM
- upgrade deps (especially React which update from 0.11.1 to 0.14.7)

**THIS PR NEEDS TO BE TESTED DEEPLY BEFORE MERGING**


aenario:
Well, this PR is too big for github so review out of context

## Problems
- [ ]  Move `require(linkedstate-mixin)`s to the top of file
- [ ]  it would be better to use `toArray` + `key` instead of `createFragment` when it works. Alternatively, we could make an helper function to optimise the use case `fn(Immutable.Map, (key, value) -> ReactElement) -> Array(ReactElement)`
- [ ] `basic_components` not sure how to do it better, but i dislike the inversion of createFactory position. Maybe we could have an helper `createFactories` ?
- [ ] `alert` module.exports = createFactory 
- [ ] `findDOMNode` vs `ref`, pick one and stick to it ( :+1:  findDOMNode)
- [ ] Add comment to weird require `date_range_picker` / `talker` / `plugin_utils`
- [ ] Dont use jquery in `file_picker`
- [ ] `createFragment` required & not used `client/app/components/mailbox_picker.coffee`
- [ ] weird filemode change on all assets
- [ ] why `manifest.json` in `vendor` ?
- [ ] I feel like the check for webpack-assets should be in config.coffee 
- [ ] server/config.coffee typo locales (langs) -> locals (local variables)


## Idea for Improvements
- [ ] Now that we can use npm modules, it would probably be better to replace `cached-transforms` by `reselect` or alternatives ( @misstick had a good one), could wait the store-to-getters refactoring
- [ ] drop jquery once we have the new editor 
- [ ] find a react-date-range-picker from NPM 

And then, you will need to rebase against @misstick recent changes :-/

aenario:
Oups, forgot to mention the main point of this review : This is awesome !

m4dz:
> This is awesome !

:kissing_cat:

m4dz:
About `basic_components`, what if we export the basic components as a factory:

```coffee
module.exports = createFactories (components...) ->
    res = {}
    for component in components
        if registry[name]
            res[component] = React.createFactory registry[name]

    return res
```

and then call requires like:

```coffee
{FieldSet, FormButtons, FormButton} = require('./basic_components')('FieldSet', 'FormButtons', 'FormButton')
```

m4dz:
> weird filemode change on all assets

does this really matters?

> why manifest.json in vendor ?

I haven't any `manifest.json` file in `vendor` locally. Are you sure about this file?

m4dz:
@aenario I left your ideas for improvements outside of this PR, but I totally agree to all of them :smile:

poupotte:
Thanks @m4dz for this PR :+1: 

I encountered some errors, I paste it on https://gist.github.com/poupotte/0629e4ffd06762024ae7 :

* When I 'check connection' for a new mail account, I receive an error. (see gist - 1)
* When I create a new account, it's created but application don't redirect to messages list. Application stay on configuration page.
* I receive regularly some errors when I navigate in my messages list (see gist - 2)
* I can't add an attachment in a new mail. If I click on 'Choose local file', I receive an error (see gist - 3). Drag and drop works.
* I can add recipient to a new mail, I receive an error (see gist - 4)

I continue my tests ...

aenario:
For the typo : it will not cause a bug, its just the wrong word for what is being passed (should be `locals`)
https://github.com/m4dz/cozy-emails/blob/d82f120cd9073ee805e7782500b8de07d839ffe7/server/config.coffee#L20-L21

:+1: for using ref everywhere.

 filemode change, doesn't really matter.

Regarding vendor, it's the whole assets folder https://github.com/m4dz/cozy-emails/tree/d82f120cd9073ee805e7782500b8de07d839ffe7/client/vendor/assets which I am not sure why you put them in vendor instead of app.

aenario:
For `basic_component`, what about 

```coffee
{FieldSet, FormButtons, FormButton} = require('./basic_components').Factories
```

It still is explicit on require side that we are taking the factories and not the classes and is simpler, and we just have to add
```coffee
export.Factories[key] = createFactory klass for key, klass of registry
```
 in basic components

m4dz:
@aenario, @poupotte: your reviews are theorically fixed. Can you pass some tests to see if everythings is OK? Thanks!

frankrousseau:
Can we merge that one?

@m4dz Can you rebase it ?

m4dz:
Done for rebase

frankrousseau:
Thank you all for that. It's a lot of work. It should make the development much cleaner. I can't wait to see tests on the client side!

PR #770, opened 2016-03-01
After reloading the page, there was nothing in the mail

misstick:


client/app/components/panel.coffee
```diff
@@ -71,6 +71,7 @@ module.exports = Panel = React.createClass
 
         # -- Generates the new message composition form
         else if @props.action is 'compose' or
+                @props.action is 'compose.edit' or
                 @props.action is 'edit' or
```
aenario: I think the 'edit' action doesn't exist anymore.

PR #772, opened 2016-03-02
Date filter are broken

misstick:
#### All these filter are broken
 - [x] filter "yesterday" 
 - [x] filter "last month" 
 - [x] filter "last week" 
 - [x] reset filter

#### Page is not updated
No messages are found after filtering but if the page is loaded, then the data is visible : 

When URL changes `messageListView`should be updated but is not becauseof `fetch` that is launched at first (loading) but not after changes.

 - [x] Fetch conversations/messages on each URL changes,
 - [x] add `keys` ont `view` to force update.

### DIFF
--- a/client/app/components/date_range_picker.coffee
+++ b/client/app/components/date_range_picker.coffee
@@ -5,7 +5,6 @@
 momentFormat     = 'DD/MM/YYYY'
 datePickerFormat = '%d/%m/%Y'
 
-
 module.exports = DateRangePicker = React.createClass
     displayName: 'DateRangePicker'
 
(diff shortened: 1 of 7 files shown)


client/app/components/date_range_picker.coffee
```diff
@@ -66,37 +64,26 @@ module.exports = DateRangePicker = React.createClass
...
+        @changeDates 'month'
 
-    presetLastMonth: ->
-        @setState
-            isActive:  true
-            startDate: moment().subtract(1, 'month').format(momentFormat)
-            endDate:   moment().format(momentFormat),
-            @filterize
+    changeDates: (type) ->
+        value = moment().subtract(1, type).startOf(type) if type
```
aenario: If we are March 16th I think this will look for message from February 1st to February 29th, where as the previous behaviour was messages from February ~16th to now.

A good middle ground might be messages from February 1st (`stubstract(1).startOf()`) to now.

aenario: ping @poupotte @kelukelu , this is actually a PO choice.

misstick: The choice is between : 
@poupotte && @kelukelu : I can resume the issue like : 
User click on 'last month' and all messages from last month only are displayed in the list
or
User click on 'last month' and all messages from last month to this day are displayed in the list

aenario: This was confirmed as expected behaviour change on IRC.

frankrousseau:
ping @aenario It's the end of the sprint. Can we merge this PR?

misstick:
@frankrousseau : This PR is not finished (see description and label); there are still bugs about filters, I have to see with Zoe where I stop the fix to make all this stable

frankrousseau:
ok

misstick:
@aenario : I've just finished!

frankrousseau:
Wow awesome! I  @aenario can you review it?

PR #779, opened 2016-03-07
Reload composeView when MessageStore fetch success

misstick:
Due to uniq `view.key` `compose` view is not as reload as before.
 - [x] Reload view when `MessageStore` changes

Regression : 
- [x] create juste one draft for a single message

### DIFF
--- a/client/app/stores/message_store.coffee
+++ b/client/app/stores/message_store.coffee
@@ -284,7 +284,7 @@ class MessageStore extends Store
 
         handle ActionTypes.RECEIVE_RAW_MESSAGE, (message) ->
             onReceiveRawMessage message
-            @emit 'fetch:success', message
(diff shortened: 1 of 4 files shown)


client/app/stores/message_store.coffee
```diff
@@ -284,7 +284,7 @@ class MessageStore extends Store
 
         handle ActionTypes.RECEIVE_RAW_MESSAGE, (message) ->
             onReceiveRawMessage message
-            @emit 'change'
+            @emit 'fetch:success', message
```
aenario: if you dont emit change, everything else will break (no refresh)

misstick:
qsdqsd

qOn Mar 8, 2016 11:25 AM, noelie wrote 
  …

PR #785, opened 2016-03-08
Fix : group messages into conversations

misstick:


client/app/components/message-list.coffee
```diff
@@ -51,7 +51,7 @@ module.exports = MessageList = React.createClass
         account   = AccountStore.getByID accountID
         role = AccountStore.getMailboxRole account, mailboxID
         settings = SettingsStore.get()
-        displayConvs = !!(settings.get('displayConversations') and
+        displayConvs = !!(settings.get('displayConversation') and
```
aenario: Feel like we might as well drop the settings.displayConversation (cc @m4dz for refactoring)

PR #787, opened 2016-03-08
FIX get next/previous conversation not message

misstick:


client/app/stores/message_store.coffee
```diff
@@ -606,13 +606,18 @@ class MessageStore extends Store
...
 
-        getMessage = ->
-            index0 = messages.toArray().findIndex (message) ->
-                if param.conversationID isnt 'undefined'
-                    return param.conversationID is message.get 'conversationID'
-                return messageID is message.get 'id'
-            index = param.transform index0
-            messages.get index
+        getMessage = =>
+            _newMessage = (index) ->
```
aenario: OK, I give up, what is this code suppose to do ?
the `getMessage` function is immediately invoked and not recursive, so seems unnecessary.
the two `index0` will overwrite each other, is that on purpose ?
`messages.toArray().findIndex` -> `message.findEntry`
what does `_newMessage` does exactly ?

aenario: On the other hand, if it fixes a bug and we actually do the big refactoring(TM), whatever ;-)

misstick: @aenario : Several messages have the same conversationID, so I check if next message isnt port of the same conversation; If you see a better way to do it, I am all ears

frankrousseau: That one is hard to understand. Moreover there is a return true without return false in other case. I think comments, explicitly handle all cases and better variable names will improve it.

PR #788, opened 2016-03-09
Clean send message contents

m4dz:
This PR continues #780 by cleaning message content before sending


server/controllers/messages.coffee
```diff
@@ -194,12 +196,30 @@ module.exports.send = (req, res, next) ->
     delete req.body.isDraft
 
     message = req.body
-    message.html = message.html.trim() if message.composeInHTML
+    if message.html
+        minifierOpts =
```
aenario: Please move static tool option at the beginning of file  or to another file. 

(Its not a big deal in this case, but will cause performance issues if moved to the render cycle in a refactoring)

aenario: Ok, I am a morron, though this was on client-side...
Request still applies but for code readability.

aenario: Feel free to open another PR, merging as is now

PR #790, opened 2016-03-10
Remove conversation configuration

misstick:
It is not a setting but natural behaviour of Email UX

### DIFF
--- a/client/app/components/conversation.coffee
+++ b/client/app/components/conversation.coffee
@@ -46,8 +46,6 @@ module.exports = React.createClass
             length = MessageStore.getConversationsLength().get conversationID
             selectedMailboxID ?= Object.keys(message.get('mailboxIDs'))[0]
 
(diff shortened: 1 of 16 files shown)


client/app/components/conversation.coffee
```diff
@@ -42,9 +42,6 @@ module.exports = React.createClass
             length = MessageStore.getConversationsLength().get conversationID
             selectedMailboxID ?= Object.keys(message.get('mailboxIDs'))[0]
 
-        displayConvs = AccountStore.hasConversationEnabled selectedMailboxID
```
aenario: I think conversation have to be disabled in trashbox otherwise we get weird behaviour. Could you check ?

misstick: @aenario : Could you explain what weird behaviour could happen? I've made some check : messages are displayed into TrashBox, I can reset it.

PR #793, opened 2016-03-14
Make routing easier

misstick:
#### Remove `panels` layout
 - [x] Remove `panel_router`(useless abstraction),
 - [x] remove `firstPanel` and `secondPanel` ,
 - [x] remove `router_panel.coffee`
 - [x] remove `component/panel.coffee`
 - [x] clean Views : add getters for 1rst level views (application, Panel)
 - [x] less `views` are listening to `Stores` changes

#### Easier `routes`
 - [x] make `Routes` easier,
 - [x] move `filters`handling into `getter/router`,
 - [x] redirection : do not redirect into `views`but with `routerActionCreator.navigate`,
 - [x] use real DOM links to change Route directly,
 - [x] clean `buildURL` parameters + change its calls,
 - [x] no `window.location.href` into `views`.

#### Clean `actionCreators`
 - [x] remove `callback` into `actionsCreator`,
 - [x] `actionsCreator` can't call other `actionsCreator`,
 - [x] remove callback to be executed after dispatch (anti-pattern)

#### More smaller `stores`
 - [x] move notifications from `layout_store` to `notification_store`
 - [x] create `selection` store

### Features concerned
Check new `routes` for :
 - [x] new Account,
 - [x] edit account,
 - [x] Message List,
 - [x] Message display,
 - [x] Message edit,
 - [x] Message reply,
 - [x] Message reply all,
 - [x] Message forward,

### Synthese
- [x] Make schemas to explain : before, after and what next?
- [x] Write guidelines to use `Router`, and `redirection` int the App

### Fix features
 - [x] Selection or mailbox `item` into left Menu,
 - [x] `Menu` can't be minimized,
 - [x] remove `fullscreen` button/mode,
 - [ ] keyboard navigation
 - [x] each message of a conversation has got its own URL


misstick:
FIXME : there still merge issue; setting application has been lost

client/app/actions/layout_action_creator.coffee
```diff
@@ -115,18 +124,52 @@ module.exports = LayoutActionCreator =
...
+        # Get Messages first
+        unless MessageStore.getCurrentConversation()
+            AppDispatcher.handleViewAction
+                type: ActionTypes.MESSAGE_FETCH_REQUEST
+            AppDispatcher.waitFor[MessageStore.dispatchToken]
 
-        # ensure the proper account is selected
-        {accountID, mailboxID} = params
-        AccountActionCreator.ensureSelected accountID, mailboxID
+        # Display Application
```
misstick: @aenario, @m4dz : I'm not really sure that is the right place to do it; waiting your advises :)

client/app/components/conversation.coffee
```diff
@@ -146,6 +146,9 @@ module.exports = React.createClass
                 messages.push(last = []) unless _.isArray(last)
                 last.push key
 
+        closeURL = Router.buildURL
```
aenario: Note : having this closeURL value not be part of state or props will cause the following bug :
- I open message list & conversation -> both are rendered, closeURL is correct
- I change the filter in message-list -> message-list is rendered but NOT conversation, closeURL doesn't change
- I click the close button -> I go back to message-list without filter

The closeURL should instead belong to AppStateStore and be fetched in getStateFromStore

client/app/components/message.coffee
```diff
@@ -22,7 +22,8 @@ LayoutActionCreator  = require '../actions/layout_action_creator'
 MessageActionCreator = require '../actions/message_action_creator'
 ContactActionCreator = require '../actions/contact_action_creator'
 
-RouterMixin           = require '../mixins/router_mixin'
+Router= require '../mixins/router_mixin'
```
aenario: If it's not a mixin, it shouldnt be in mixins folder, may be router.coffee at client's root ?

client/app/stores/message_store.coffee
```diff
@@ -360,18 +361,31 @@ class MessageStore extends Store
...
+        handle ActionTypes.MESSAGE_FETCH_REQUEST, ->
+            return if not (url = @getNextUrl()) or @isFetching()
+
             # There may be more than one concurrent fetching request
             # so we use a counter instead of a boolean
-            _fetching++
-            @emit 'change'
+            mailboxID = AccountStore.getSelectedMailbox()?.get 'id'
+            ts = Date.now()
+            XHRUtils.fetchMessagesByFolder url, (err, rawMsg) ->
```
aenario: This logic shouldn't be in the store, stores are supposed to be dumb. 

I think it does belong to the ActionCreator.

client/app/components/application.coffee
```diff
@@ -40,153 +30,58 @@ MessageActionCreator = require '../actions/message_action_creator'
...
-        # TODO: Improve the way we display a loader when app isn't ready
-        layout = @props.router.current
-        return div null, t "app loading" unless layout?
-
-        disposition = LayoutStore.getDisposition()
-        isCompact   = LayoutStore.getListModeCompact()
-        fullscreen  = LayoutStore.isPreviewFullscreen()
-        previewSize = LayoutStore.getPreviewSize()
+    getDefaultProps: ->
+        ApplicationGetters.getProps 'application'
```
aenario: We need to decide a guideline for getters : 
In my humble opinion : 
`getDefaultProps` and `PropTypes` belongs to the view
`getStateFromStore` should be explicit :
 ```coffee
getStateFromStore: ->
     stuff1 : SomeGetter.getStuff1()
     stuff2 : SomeGetter.getStuff2()
```
`ApplicationGetters.getMenuProps() ` rather than `ApplicationGetters.getProps('menu')` (which by the way should go away as Menu is a Controller)

File: client/app/components/conversation.coffee (hunk 2 above)
Comment (misstick) [inline client/app/components/conversation.coffee]: @aenario : I'm OK with you but you're going to fast, it's still in progress :)

File: client/app/components/message.coffee (hunk 3 above)
Comment (misstick) [inline client/app/components/message.coffee]: @aenario : same

File: client/app/stores/message_store.coffee (hunk 4 above)
Comment (misstick) [inline client/app/stores/message_store.coffee]: @aenario : same

File: client/app/actions/layout_action_creator.coffee (hunk 1 above)
Comment (aenario) [inline client/app/actions/layout_action_creator.coffee]: This will be called every time the user comes back to mesage-list from elsewhere in the app, wont it ?

Why do you want to remove it from initialize.coffee ?

File: client/app/router.coffee [hunk 6]
```diff
@@ -1,96 +1,156 @@
...
+        '/*'                                    : 'messageList'
+        'account/new'                           : 'accountEdit'
+        'account/:accountID/config/:tab'        : 'accountNew'
+        'search/?q=:search'                     : 'search'
+        ':messageID'                            : 'messageShow'
+        ':messageID/edit'                       : 'messageEdit'
+        'new'                                   : 'messageNew'
+        ':messageID/forward'                    : 'messageForward'
+        ':messageID/reply'                      : 'messageReply'
+        ':messageID/reply-all'                  : 'messageReplyAll'
```
m4dz: I'm not sure about those routes: is your goals to mimic the https://github.com/cozy/cozy-emails/pull/783/files#diff-f43c9098b0e1615b73bef5226fbe582f definition at final form, or is this just transitional?

client/app/components/application.coffee
m4dz: I agree to the concepts exposed by @aenario but wait to see the `ApplicationGetter` part to be sure about the final form.

m4dz:
This is definitely a very good proposal for the new Router structure. For me, it push the app in the right direction, good job!

client/app/components/compose.coffee
```diff
@@ -441,7 +441,7 @@ module.exports = Compose = React.createClass
             # to save temporary info
             unless @state.id
                 @state.id = message.id
-                Router.redirect
+                window.location.href = RouterGetter.getURL
```
misstick: It is a temporary FIX; this should be moved into `action_creator` files, and use `navigate`from `Backbone.Router`

misstick:
I should create a `RouterStore` and a `RouterActionCreator` from `RouterGetter`; actually since last commit application is broken

client/app/initialize.coffee
```diff
@@ -129,17 +127,7 @@ document.addEventListener 'DOMContentLoaded', ->
...
-
-        # Binds the router and flux to the React application
-        Application = React.createFactory require './components/application'
-        application = Application router: @router
-        rootNode    = document.querySelector '[role=application]'
-        ReactDOM.render application, rootNode
-
-        # Starts the application by initializing the router
-        Backbone.history.start()
+        window.router = new Router()
```
misstick: FIXME : replace `window.router` reference by `RouterGetter`

client/app/stores/message_store.coffee
misstick: @aenario : `MessageActionCreator` should be more appropriate than `accountActionCreator`?

client/app/components/toolbar_messageslist_actions.coffee
```diff
@@ -86,17 +82,12 @@ module.exports = ActionsToolbarMessagesList = React.createClass
...
-                @redirect
-                    direction: 'second',
-                    action: 'conversation',
-                    parameters:
-                        messageID: nextConversation.get('id')
-                        conversationID: nextConversation.get('conversationID')
+            # FIXME : move this into MessageActionCreator
+            # after deleting complete
+            # Goto to next conversation
+            window.location.href = RouterGetter.getURL
```
misstick: FIXME : use `RouterActionCreator` to dispatch an action that will handle redirection.

client/app/components/application.coffee
misstick: @aenario :
`ActionCreator` are the `controller` of all the `states` of the Application also known as `Stores`.
After dispatching an `Action` we can dispatch a `success` or `error` action to update application (ie. fetching message) and :
 - make redirections (and remove it from `views`),
 - update data (from this redirection if needed).
 
Handling that kind of action should prevent us to `listenTo` `Stores`; that's why I think we can remove all `StoreWatch` logic.

What do you think?

misstick: @aenario : Are you still agree with this idea, because I AM :) It should be easier to implement, before adding a `redux` logic (or other stuff)

client/app/actions/router_action_creator.coffee
```diff
@@ -13,20 +13,47 @@ RouterActionCreator =
...
+
+    navigate: (params={}) ->
+        {url} = params
+        url ?= RouterStore.getURL params
+
+        console.log 'navigate', params
+        if url
             # Update URL && context
+            router = RouterStore.getRouter()
             router.navigate url, update: true
```
misstick: !!! use `dispatcher` to change `URL`

frankrousseau:
@misstick Is it possible to merge that PR and open a new one with the remaining of the todo-list?

misstick:
@frankrousseau : no it is not finished yet, and it need to be reviewed

frankrousseau:
Yep but I don't like huge PR. That one will be close to non reviewable because of its size.

misstick:
@frankrousseau : yes it can't be split in other PR : this feature has too much consequences to be smaller

m4dz:
@frankrousseau this PR concerns the core routing that also impacts stores, and I can't see a clean way to split it without side-effects. We made a meeting with @poupotte and @aenario this morning, we'll see what we can do to fix it :).

frankrousseau:
Ok @misstick and @m4dz

frankrousseau:
Thx for the info update.

m4dz:
@frankrousseau no problem :). We'll merge it after pre-review (from me and @aenario) tomorrow, and then use it as a solid base structure for the rest of the app. It'll let us to see where app is broken and where it is stable, and the best practices to apply. Plus, we'll be able to work on dedicated parts each of us at the same time. It's on the right way!

client/app/getters/router.coffee
```diff
@@ -21,6 +21,21 @@ class RouteGetter
...
         RouterStore.getAction()
 
+    isReply: ->
+        action = @getAction()
+        action isnt 'message.edit'
+
+    isEditable: ->
+        action = @getAction()
+        editables = [
+            'message.new',
```
misstick: TODO : use `ComposeActions ` here

aenario:
Could you try ton run 
```bash
./node_modules/.bin/coffee-jshint --options browser,node,undef,unused client/app/**/*.coffee --globals t  > lint.log
```

Most of them are not your fault, but i see some in account_action_creator, which will crash the client on mailbox manipulation.

aenario:
REMOVE_ACCOUNT_XXX actions are comented but used.

aenario:
#MAYBELATER

There is quite a few unused require in stores & action_creators (for instance the layout action creator requires but do not use XHRUtils, SocketUtils, MessageStore, AccountActionCreator & MessageActionCreator). 

This is a sign that you've done a good job cleaning the mess in the dependencies graph :+1:

aenario:
router_action_creator.coffee#L28
Prefer for key, value of params instead of _.each

aenario:
#MAYBELATER

router_action_creator.coffee#L46
French comment makes @clochix happy but we try to be welcoming to people who dont have the chance to speak french :-p (will be gone with #FIXME anyway)

aenario:
components/application.coffee
I would move the `Router.getLayoutSettings` to getStateFrom Stores
and `className = "" + "..." +` to the render method`
in order to get rid of the whole `props` part.

aenario:
panel.coffee  this file is not used anymore ( :tada:  thanks !) kill it :skull:

aenario:
getters/routers.coffee
You have a lot of if/unless (variable = condition) where you dont use the variable after.
isFlags takes one argument, you use it with two

getters in general : 
Not sure if they should be a class or map of function, will probably depends on how we do memoization.

aenario:
getters/selection
Making it two functions instead of getProps avoid the strange _.extend in message_list

aenario:
Could you explain your changes in store_watch_mixin ? 

I find the new approach harder to read with the mix in getStateFromStores & getInitialState.

aenario:
missing require AccountStore in NotificationStore

aenario:
I see you added some logic in XHRUtils, IMHO, it should go back to action_creator.

aenario:
#MAYBELATER

I think we should move all the _current and _selected variables to the routerStore, and get rid of `LayoutActionCreator.saveMessage` but this can be part of another PR. The current state is already much cleaner !

frankrousseau:
\o/

misstick:
I thought that moment would never happen!
https://c1.staticflickr.com/5/4025/4488876837_8d3da2423a_z.jpg

PR #796, opened 2016-03-15
Issues on messages basic feature

misstick:


client/app/components/compose.coffee
```diff
@@ -127,18 +127,14 @@ module.exports = Compose = React.createClass
...
+        # Save client changes
+        @state.lastUpdate = new Date().toISOString()
 
     hasChanged: (props, state) ->
-        (props.lastUpdate and props.lastUpdate isnt state.date) or false
+        @state.lastUpdate isnt @state.date
 
     resetChange: ->
-        delete @props.lastUpdate
+        @state.lastUpdate isnt @state.date
```
misstick: rebase issue ? it should be `@state.lastUpdate = @state.date`

aenario: Weird... Could you make a PR with the fix :-)

PR #798, opened 2016-03-17
Client test stack

m4dz:
This PR is an **experiment** at this time. Do _not merge_ before we take some decisions about its concerns and behaviors.

The PR tends to prove that client non-view components like Stores and Flux patterns can be unit-tested headlessly to ensure the application fondation is stable enough.


nono:
@m4dz why did you choose to try tape?

frankrousseau:
Please don't include Casper.

frankrousseau:
And same as @nono please don't change mocha.

m4dz:
Tape seems to be more and more pushed to the front by the comunity, and it have several advantages:

- fast (faster than mocha)
- reliable
- do not pollute global environement: every test is sandboxed in closures
- efficient assertions

But it's just a test for the moment, I don't want to change some habits before we discuss together about the need / want to change the runner.

frankrousseau:
Casper was already present, sorry. Can you remove it?

frankrousseau:
Please, we already have numerous technologies. We are satisfied with Mocha. It's not the time to add another technology to our stack.

m4dz:
@frankrousseau Yep, I'll remove Casper to let only unit-tests at this time. We probably want to make some functional tests to validate scenarios later, and then check to reintroduce it, but it's a subject for another PR, later (mean: after refctor :wink:)

About the choice of the runner, I'm totally open. This one is just a test, we can switch back to mocha if we still find it more efficient.

nono:
@m4dz If you want to test runners, I think [painless](https://github.com/taylorhakes/painless) has some interesting features. But I agree with Frank, mocha works pretty well and I don't see a good reason to change for the moment.

client/test/layout_store.spec.js
```diff
@@ -0,0 +1,60 @@
...
+            assert.equal(layoutStore.getDisposition(), Dispositions.COL)
+        })
+
+        it('Switch to ROW: DISPOSITION is ROW', function (done) {
+            layoutStore.on('change', function () {
+                assert.equal(layoutStore.getDisposition(), Dispositions.ROW)
+                done()
+            })
+
+            dispatcher.dispatch({
```
aenario: Store change should be synchronous but i am not sure if the tests should reflect that.

client/test/utils/specs_dispatcher.js
```diff
@@ -0,0 +1,16 @@
+var TestDispatcher = function TestDispatcher() {
```
aenario: May be the mock can just use client/lib/flux/dispatcher/dispatcher.coffee which is a self-contained reusable class and would be closer to appDispatcher behaviour when we want to test more complex behaviour such  as waitFor.

frankrousseau:
@m4dz Could you fix @aenario remarks?

m4dz:
This one touch to the Dispatcher, and we (probably) won't keep it as it after the refactor. I made it for demo purposes first, so 2 solutions for me:

1. we fix @aenario remarks but certainly trash them later
2. we merge it in this state and evolves it after the refactor

@aenario, @frankrousseau any prefs?

frankrousseau:
I prefer we merge it. Putting tests in this app is important.

m4dz:
> I prefer we merge it. Putting tests in this app is important.

I agree: in this state, app is ready to run tests, we _just_ need to write them :smile:. @aenario, good for you?

PR #804, opened 2016-03-25
Fix compose features

misstick:
- [x] Fetching after save create a conflict : 2 dispatch on the same Store
 - [x] `ComposeInHTML` options was disabled becauseof bugs
 - [x] Disable `ComposeInHTML` because it is unstable
 - [x] remove buggy focus
 - [x] improve `@hasChanged`
 - [x] fetch after save into `messageACtionCreator`
 - [x] when I send a message, its `body` is empty

### DIFF
--- a/client/app/components/compose.coffee
+++ b/client/app/components/compose.coffee
@@ -65,7 +65,7 @@ module.exports = Compose = React.createClass
         # Get Reply message
         if _.isString props.inReplyTo
             id = props.inReplyTo
-            if (message = MessageStore.getByID id) and message.size
+            if (message = MessageStore.getByID id)?.size
                 message.set 'id', id
                 props.inReplyTo = message
         MessageUtils.createBasicMessage props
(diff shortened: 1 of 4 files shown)


frankrousseau:
@m4dz Could you review that one?

client/app/components/compose.coffee
```diff
@@ -113,109 +108,33 @@ module.exports = Compose = React.createClass
...
 
-        # Each state:change do not send data to server
-        # update date for client modifications
-        @state.lastUpdate = new Date().toISOString()
 
     hasChanged: (props, state) ->
-        @state.lastUpdate isnt @state.date
+        _diff = _.filter @state, (value, key) =>
+            @_oldState[key] isnt value
+        _diff?.length
```
m4dz: I think we can use `_.isEqual` to test changes, nope?

misstick: It doesnt send the same result :(

m4dz:
Thanks! I made another PR to disable the compose module :wink:.

PR #807, opened 2016-03-30
Disable Compose feature

m4dz:
The PR temporary disable Compose actions and paths and let the app in a read-only mode. This PR will probably be reverted when doing the big refactor.

### DIFF
--- a/client/app/components/menu.coffee
+++ b/client/app/components/menu.coffee
@@ -87,9 +87,6 @@ module.exports = Menu = React.createClass
             role: 'menubar'
             'aria-expanded': @state.isDrawerExpanded,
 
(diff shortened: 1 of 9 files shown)


frankrousseau:
One of the goal is to make the big refactor faster. So, please do not revert it while doing it.

m4dz:
Yup yup, I misexplained it. I mean: I don't just comment the code, but really remove it to prevent routes to accidentally activate the compose mode (e.g. if you point your browser to /compose). So I suggested to keep this PR as a reference when we are ready the add compose again to easily find what I removed (but I don't want to revert it as it, for sure).

frankrousseau:
Thank you @m4dz

PR #814, opened 2016-04-06
Improve messages workflow

misstick:
## Improve features
 - [x] upgrade `unread messages counter` into menu as fast as possible,
 - [x] active realtime events,
 - [x] fix `nextURL` setting,
 - [x] do not fetch all messages to get `conversationLength`,
 - [x] fetch all messages from conversation when a message is displayed,
 - [x] show only current message on 2nd panel,
 - [x] reset `currentID` when Route changes,
 - [x] clean messageList dead code,
 - [x] fix keyboard navigation for `top` and `bottom` keys,
 - [x] fix `filter` navigation,
 - [x] fix checkbox `selection`,

<strike> - [ ] `selection` menu: remove tools submenu to display simple buttons </strike>

## Remove features
 - [x] `search bar` is useless (search is broken),
 - [x] `filter toolbar`,
 - [x] add mailbox `unread`,
 - [x] add mailbox `important`,
 - [x] compose was re-added but should disappear, see: https://github.com/cozy/cozy-emails/pull/807


client/app/components/conversation.coffee
```diff
@@ -106,7 +103,7 @@ module.exports = React.createClass
 
             section
                 ref: 'scrollable',
-                    @state.messages.map @renderMessage
+                    @state.conversation.map @renderMessage
 
     _initScroll: ->
         if not (scrollable = ReactDOM.findDOMNode @refs.scrollable) or scrollable.scrollTop
```
frankrousseau: Are you sure that this line is < 80 chars?

frankrousseau:
I didn't notice before but it looks like you don't follow the 80 chars guideline. Could you stick to it? It makes the text size standard, force to write more concise code and match with what we do since the beginning.

frankrousseau:
Too much PR open for the end of the sprint. Please merge or close them on monday.

misstick:
@m4dz : I've juste finished the fiw because of merge; you can review this PR when ever you want :]

client/app/components/search_bar.coffee
```diff
@@ -66,7 +66,7 @@ module.exports = GlobalSearchBar = React.createClass
         .toOrderedMap()
         .set 'all', t 'search all accounts'
 
-        accountID = AccountStore.getSelected()?.get('id') or 'all'
+        accountID = AccountStore.getAccountID() or 'all'
```
m4dz: Should use a getter here, nope?

client/app/stores/account_store.coffee
```diff
@@ -379,62 +252,9 @@ class AccountStore extends Store
...
-        account.imapTLS = false
-        account.accountType = 'IMAP'
-        account.favoriteMailboxes = null
-        return Immutable.Map account
-
-
 
 
-module.exports = new AccountStore()
+module.exports = _self = new AccountStore()
```
m4dz: Why storing instance in `_self`?

m4dz:
I merge as it, maybe we need to think about:
- using Getter in component and never call Stores directly
- Use dedicated getters for stores like MessagesStores, and not let RouterGetter serves contents that doesn't rely directly on it (like `getMessagesList`, which should be in a MessagesGetter that can ask RouterSTore for the current `mailboxID`)

We also need to add a lot of comments in the whole app: each time I review, it's deeply hard to understand who does what, how, and what have changed :disappointed:…

misstick:
@m4dz : 
 - " using Getter in component and never call Stores directly" : this should be done a little each time 
 - for getter, we could talk about this; to my mind, having only one getter si enought but its name should be change in `getters/application` ie.

client/app/stores/account_store.coffee
misstick: Sometime references to `this` with `@` doesnt work...

PR #824, opened 2016-04-15
Before setting action, check if any account exists

misstick:


misstick:
@m4dz : could review and merge as soon as you can this PR?

client/app/stores/router_store.coffee
```diff
@@ -186,7 +186,12 @@ class RouterStore extends Store
             @emit 'change'
 
         handle ActionTypes.ROUTE_CHANGE, (value) ->
-            _action = value
+            # We cant display any informations
+            # without accounts
+            unless AccountStore.getAll()?.size
+                _action = AccountActions.CREATE
```
m4dz: As long as `ROUTE_CHANGE` become the default event dispacthed by the router on url updates, are we sure that the requested action is `AccountActions.CREATE` or should we use the `value.action` passed by the router instead?

misstick: this patch was done according to this only one `action` dispatched. The tricky case is when user has no account, and the url is `http://localhost:9125/` where `value.action` is equal to `AccountActions.SHOW_ALL`. In this case no messages could be found. Changing the param here make the redirection : I didn't fount a better place to do this.

frankrousseau:
Too much PR open for the end of the sprint. Please merge or close them on monday.

PR #823, opened 2016-04-13
Move Router dispatch to uniq ROUTE_CHANGE

m4dz:
This PR clean routing and actionsCreators.

- [x] Make initialize runtime more decoupled and simple
- [x] Clean dispatcher
- [x] Use a single `ROUTE_CHANGE` event on URL updates
- [x] Redirect root URL to first account's INBOX
- ~~Clean ActionsCreators~~

PR #826, opened 2016-04-20
[FIX] merge conflicts between #823 and #814

misstick:
- [x] when no account found, do not redirect to `#mailbox/undefined` but `account/new`,
 - [x] when refreshing `#mailbox/:mailboxID/:messageID` : email and list are not visible,
 - [x] FIX `_accounts.get(...) is undefined`,

<strike> - [ ] FIX `error_manager.coffee:29 POST http://localhost:9125/activity  413 (Payload Too Large)`,</strike>
 <strike>- [ ] Keyboard navigation is broken</strike>

### DIFF
--- a/client/app/app_dispatcher.coffee
+++ b/client/app/app_dispatcher.coffee
@@ -8,7 +8,6 @@ class AppDispatcher extends Dispatcher
 
     dispatch: (action) ->
         window.cozyMails.logAction action
-        window.cozyMails.customEvent PayloadSources.SERVER_ACTION, action
         super {action}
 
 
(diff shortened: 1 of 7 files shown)


misstick:
@m4dz : is `error_manager.coffee:29 POST http://localhost:9125/activity 413 (Payload Too Large)` normal? why?

misstick:
@m4dz can you explain why Keyboard navigation is broken? and why I shouldnt FIX it?

m4dz:
About unfixed bugs:
1. "Payload too large" is an error that occurs when React try to POST `console.debug logs` to the server logger. It never occurs in production mode, as the React `console.debug` calls are triggered when app run in `__DEV__` config, so we can safely ignore them
2. Keyboard navigation is broken because it is handled via a plugin, and we decided to trash the actuel plugins stack, and 1/ reintroduce it maybe later if needed (with a dedicated API); 2/ add keyboard events supports natively in the app. So feel free to let it unfixed at this time :smiley:

m4dz:
@misstick can i let you rebase it on development? Thanks!

PR #828, opened 2016-04-22
[Refactor] Prefer modules to globals

misstick:
- [x] remove `window.cozyMails`,
 - [x] remove `window.__DEV__`,
 - [x] remove `window.settings`,
 - [x] remove `window.router`


client/app/components/account_config.coffee
```diff
@@ -82,11 +82,12 @@ module.exports = React.createClass
         @setState nextstate
 
     render: ->
+        title = if @isNew() then 'account new' else 'account edit'
         Container
             id: 'mailbox-config'
             expand: true,
 
-            Title text: t if @isNew() then 'account new' else 'account edit'
+            Title text: t "#{title}"
```
m4dz: Can be simply: `t title`

client/app/components/menu.coffee
```diff
@@ -137,8 +139,10 @@ module.exports = Menu = React.createClass
...
         }
 
+        mailboxes = @state.mailboxes.toArray()
+        className = classNames active: props.isSelected
         div
-            className: (className = classNames active: props.isSelected),
+            className: className,
             key: props.key,
             div className: 'account-title',
                 a
```
m4dz: All indent below this point seems wrong, nope?

client/app/stores/account_store.coffee
```diff
@@ -118,6 +118,7 @@ class AccountStore extends Store
     __bindHandlers: (handle) ->
 
         handle ActionTypes.ROUTE_CHANGE, ({accountID, mailboxID, action, tab}) ->
+            console.log 'SAVE', {accountID, mailboxID, action, tab}
```
m4dz: Maybe useless, nope?

client/app/stores/refreshes_store.coffee
```diff
@@ -4,6 +4,8 @@ Store = require '../libs/flux/store/store'
 
 {ActionTypes} = require '../constants/app_constants'
 
+ApiUtils = require '../utils/api_utils'
```
m4dz: Seems not in use in the module

m4dz:
Seems a good direction for me, just 2 questions:
1. I'm not especially fan that the flux lib exposes a singleton (especially when we'll want to do tests on dispatchers), maybe let application create a new dispatcher at initialization and then require it should be a better solution?
2. The only global env that seems legit for me is the `__DEV__` one. Not sure replacing it by an `isDev()` util is really a KISS way…

client/app/stores/refreshes_store.coffee
misstick: 👍  Yes it is, but code was commented

misstick:
@m4dz : I'm OK for the singleton, but as it is a temporary situation, I think we could wait to change this when add a framework based on Flux

m4dz:
If it won't break our (poor) tests stack, go to let it as it and: 1/ fix it later if needed or ; 2/ trash it when going to Redux. Thanks!

PR #839, opened 2016-05-06
[feat] Account creation wizard

m4dz:
This PR introduces a new component to manage Account creation and Settings:

- [x] Create new components architecture
- [x] Only use a global state and delegate props to sub-components
- [x] Add basic layout styles to wizard view
- [x] Bind auto-discover (and auto-expanded to advanced settings)
- [x] Feedback on autodiscover errors
- [x] Add 2FA alert on Gmail accounts
- [x] Save account in Stores
- ~~Unit test view on custom behaviors (such as magic inputs reflection)~~
- ~~Test consistent component state~~
- ~~Clean old account creator~~


client/app/components/application.coffee
```diff
@@ -41,7 +42,14 @@ Application = React.createClass
...
         TooltipRefesherMixin
-        StoreWatchMixin [SettingsStore, RefreshesStore, RouterStore, MessageStore, LayoutStore, AccountStore]
+        StoreWatchMixin [
+            SettingsStore
+            RefreshesStore
+            RouterStore
+            MessageStore
+            LayoutStore
+            AccountStore
+        ]
```
misstick: `routerStore` contains all status of the application so that `MessageStore `, `LayoutStore `, or `AccountStore ` shouldn't be listened to. Can't you remove listener? If not why?

m4dz: I've just review indent here but: I think any top-level component (aka: a component that's aware of stores changes) _should_ listen to the _main store_ (aka RouterStore) only, except if it needs to access specific data (e.g. Notifications from the NotificationStore).

Note that when we'll switch to Redux, this question won't remains as redux exposes only one store that trigger the change event :wink:

client/app/components/accounts/wizard/creation.cjsx
```diff
@@ -1,35 +1,75 @@
...
+        # `isChecking` need discover to be explicitely set to `true` (request
+        # in flight)
+        _.defaults state,
+            isChecking:     discover is true
+            isDiscoverable: true
 
 
     render: ->
+        # `required`: needs email and password filled to enable submit button
+        required = not(_.isEmpty(@state.email) or _.isEmpty(@state.password))
```
misstick: You should use `@state.required` instead; do not define data into `render`.

client/app/components/accounts/wizard/creation.cjsx
```diff
@@ -45,56 +85,81 @@ module.exports = AccountWizardCreation = React.createClass
                             value={@state.password}
                             onChange={_.partial @updateState, 'password'} />
 
-                <Servers expanded={!@state.isDiscoverable}
+                {<p className="alert">{t('account wizard is not discoverable')}</p> unless @state.isDiscoverable}
```
misstick: !! Be careful of line length

misstick:
Dont forget ton hide custom configuration fields!

m4dz:
Ready-to-review! Still lacks:
- a better layout design
- a clever button to force closing modal instead of waiting the timeout, and a proper redirect
- dead code cleaning
- tests

But I think we can do them in another PRs :smile:.

Thanks!

poupotte:
In my mind, we'll be able to merge this PR when redirection to 'account/new' will work. 
Indeed, creation of a second account isn't possible for the moment.

Better layout design, dead code cleaning and tests can be merge in another PR.

client/app/stores/requests_in_flight_store.coffee
```diff
@@ -0,0 +1,35 @@
+Immutable = require 'immutable'
```
misstick: @m4dz : this file name is generic but used only for discovering; are you sure you need this? cant you move this into `accountStore`, or `routerStore` instead?

misstick:
@m4dz : shouldnt you remove old components? Is this available to edit accounts?

client/app/components/accounts/wizard/creation.cjsx
```diff
@@ -0,0 +1,186 @@
...
+
+    displayName: 'AccountWizardCreation'
+
+    mixins: [
+        StoreWatchMixin [RequestsInFlightStore]
+        AccountMixin
+    ]
+
+
+    # Build state form RequestsInFlightStore:
```
misstick: @m4dz : to update all application, you should listen to `routerStore`instead. I still don't think that you should create a store to handle success/errors : use `routerStore` for that stuff (because its all about URL stuff)

client/app/components/accounts/wizard/creation.cjsx
```diff
@@ -0,0 +1,186 @@
...
+    'CREATE_FAILED':   'CREATE_FAILED'
+
+
+module.exports = AccountWizardCreation = React.createClass
+
+    displayName: 'AccountWizardCreation'
+
+    mixins: [
+        StoreWatchMixin [RequestsInFlightStore]
+        AccountMixin
```
misstick: `mixins` pattern is for several uses but it is obbiously used only once : use functions into that file instead

client/app/components/account_config.coffee
```diff
@@ -46,8 +46,8 @@ module.exports = React.createClass
         nstate.isWaiting         = RouterStore.isWaiting()
```
misstick: It seems that `_newAccountWaiting` has been replaced by `RouterGetter.getRequestStatus`; please remove old methods that you have replaced (for good! :) such as `isWaiting()`, `isChecking()`, `_newAccountWaiting`, `_newAccountChecking`, `_serverAccountErrorByField` etc.

client/app/components/account_config_main.coffee
```diff
@@ -249,7 +249,7 @@ module.exports = AccountConfigMain = React.createClass
                 # Check for errors before.
                 FormButton
                     class: 'action-check'
-                    spinner: @props.checking
+                    # spinner: @props.checking
```
misstick: Is this useless? If yes please remove it

client/app/components/accounts/wizard/creation.cjsx
```diff
@@ -0,0 +1,186 @@
...
+    ]
+
+
+    # Build state form RequestsInFlightStore:
+    # - is an autodiscover request in action, or a provider available?
+    getStateFromStores: ->
+        state       = {}
```
misstick: cant you move all these test into `getters`?

misstick:
Nice work! You should remove old code that seems to conflict with the new pattern; keep in mind that `mixins` are `anti-pattern` and all data logic should be moved as much as possible into getters. 👍 Keep focus, and everything should be OK. I'm gonna right these things into documentation (such as `mixins` are anti-pattern)

client/app/components/accounts/wizard/creation.cjsx
```diff
@@ -0,0 +1,186 @@
...
+                         {..._.omit @state, 'isOAuth', 'isDiscoverable', 'isBusy'} />
+
+                <footer>
+                    <nav>
+                        {<button type="submit"
+                                onClick={@create}
+                                disabled={not @state.enableSubmit}>
+                            {t('account wizard creation save')}
+                        </button> unless @state.success}
+                        {<a href="#" className="alert success">
```
misstick: what is happening when `onClick`? It should have a `onClick` or real `URL` into `href`attribute.
It it isnt a link : use button?

client/app/components/application.coffee
```diff
@@ -139,18 +123,22 @@ module.exports = React.createClass
...
+                                subject         : message?.get 'subject'
+                                conversation    : @state.conversation
+                        else
+                            section
+                                'key'          : 'placeholder'
+                                'aria-expanded': false
+
+            if @state.action is AccountActions.CREATE
+                div role: 'complementary',
+                    AccountWizardCreation()
```
misstick: You should use props to define your component; moving success, errors message/status here instead of getting them as `states` into `accountComponent`

client/app/components/application.coffee
```diff
@@ -105,26 +106,9 @@ module.exports = React.createClass
...
-                            action               : @state.action
-                            message              : message
-                            inReplyTo            : RouterGetter.getReplyMessage @state.messageID
-                            settings             : SettingsStore.get()
-                            account              : RouterGetter.getAccount()
-
-                    else
-                        div
-                            className: 'panels'
+                    div
```
misstick: If I understand completely `new account` behaviour, default mailbox should appear below `new account` modal; BUT there is an issue here because emails from this mailbox are not leaded because `RouterStore._action` is not `Mailbox.SHOW_ALL`

misstick:
m4dz: I will merge this PR into `developement` but we will have to finish this review tomorrow

client/app/stores/requests_in_flight_store.coffee
m4dz: This is need to keep tracks of all requests performed in background. We choose to use it previously with @aenario when designing the new architecture. It's also a needed component for a stateless version that'll comes next. For all of this, I've choose to introduce it now. Do it appeal to you, or should we move to another pattern?

m4dz:
> @m4dz : shouldnt you remove old components? Is this available to edit accounts?

Yep: it's in my scope for the current sprint. If we have some extra-time, we'll probably can create the edit component that reuse the `<Servers>` one.

client/app/components/accounts/wizard/creation.cjsx
m4dz: In this case, I think not: background operations aren't rely to URL, and we need to be aware of them in top-level components. I think it's out of RouterStore's responsability area to keep information about this contexts.

client/app/components/accounts/wizard/creation.cjsx
m4dz: Are _mixins_ as bad as we suggest ;)? In this case, I'm curious about some feedback: we'lle have two top level components: _AccountCreation_ and _AccountEdit_. Both exposes a `<Server>` component and manage an _Account_ data from _AccountStore_. They 're similar but not strictly the sames. We need a way to share some logics, taht are pure data parsing (and that never _get_ some information from a store nor anywhere). Plus, we need the same parsing logic in some _ActionCreator_. I think _mixins_ are trhe best way to handle this use case. What's your suggestion?

client/app/components/account_config.coffee
m4dz: OK, thanks :+1:!

client/app/components/account_config_main.coffee
m4dz: In fact, it was just a lazy comment, but the whole _account_config_main_ file is useless and will be removed :smile:

client/app/components/accounts/wizard/creation.cjsx
m4dz: Why? Aren't state responsible of maintaining data structure inside a top level component (like this wizard)?

client/app/components/accounts/wizard/creation.cjsx
m4dz: Need to talk about this one: a route change should occurs when we clicking in parallel of a timer that close the modal to redirect the user inside the mailbox. I miss informations about the _how_ :smile:

client/app/components/application.coffee
m4dz: Not agree: we defines in architecture design sprint that _Application_ component and top-level components (such as _Menu_, _MailboxList_, _Conversation_, _Compose_, _Settings_…) manage their own states (for many reasons, like to prevent useless cascading re-renders). This wizard_ is just as the _Setting_ component: it's autonomous and can handle its own state.

client/app/components/application.coffee
m4dz: That's right, see my comment about last redirection that doesn't work as well as wanted

client/app/stores/router_store.coffee
```diff
@@ -464,6 +468,11 @@ class RouterStore extends Store
         _URI = "#{mailboxID}#{query}"
 
 
+    _gotoAccountMailbox = (account) ->
+        _newAccountWaiting = false
+        _checkForNoMailbox account
+        _action = MessageActions.SHOW_ALL
```
misstick: @m4dz: you should specify `mailboxID` too

PR #850, opened 2016-05-16
Add Lodash to client dependencies and use includes instead of contains

frankrousseau:


misstick:
What the aim of this PR : FIX the build? Will you FIX all this https://github.com/lodash/lodash/wiki/Deprecations or shall we do it in an other PR?

frankrousseau:
@misstick the goal is to fix the build.

misstick:
@frankrousseau : Is you PR finished and ready to merge? What about other `may break the build` cases listed about `lodash v4.0.0` depreciated ?

frankrousseau:
I think it's ready to merge, I'm going to check if there are other problems.

frankrousseau:
Without this PR, the app doesn't work at all.

frankrousseau:
I searched for deprecated function calls and I didn't find one except the ones fixed by this PR.

misstick:
👍

PR #854, opened 2016-05-19
[fix] account creation redirect

m4dz:
This PR fix some behaviors like cancel button and form submission. It also introduces a proper redirect when creating a new accont.


misstick:
👍

PR #856, opened 2016-05-19
[fix] flickery advanced params opening at submission

m4dz:
This PR fix an unwanted behavior that open advanced parameters when re-triggering an AUTH_CHECK request on autodiscovered account.


misstick:
@m4dz : it seems that a test is broken (timeout) : 
````  4 passing (43s)
  1 failing
  1) Server tests Account creation When I post a new account to /accounts:
     Error: timeout of 12000ms exceeded
```

PR #857, opened 2016-05-19
Do not scroll everytime into components

misstick:


m4dz:
Tests seems to fail w/ node 0.12, but this is not related to this one. Thanks :tada:!

PR #860, opened 2016-05-20
Clean from #839 comments

m4dz:
This PR is in continuity of #839, base on the code review from @misstick. It introduces:
- move state logics into getters
- rename the so-badly-named `RequestsInFlightStore`
- disable legacy code for requests states and errors

Concerning this last, I keep code commented at this time as long as a dedicated PR for dead-code removing will become soon.

@misstick, I have some points that I won't address:
- The StoreWatchMixin in the _Creation_ component: I assume a modal isn't directly relied to the application and won't inherit of all its states. Plus, I listen the _RequestsStore_ rather than the RouterStore as long all its events loops comes from background requests only. I assume we improve this when we'll implement the real URL to modal (in query params) logic, as discussed
- There's always a mixin for code sharing between the _Creation_ component, the upcoming _Settings/Edit_ component and some _action creators_. If you prefer to keep it isolated in another way (a lib? a getter (but it only _parse_ and never _get_ data)?), let me know
- I haven't implemented the query param routing for modals, I prefer to keep it in a dedicated PR (and it isn't an uncontrollable urge)

**EDIT**: as discussed in review, about architecture questions:
- Using a `StoreWatchMixin` is OK for top-level components only (aka: the application's main view, and the various modals), at the condition they use a proper `key` on component that is uniq for same behaviors (settings, account-creation, draft, etc)
- Mixin should be avoid, except if what we want is an inherited behavior accross various components. Here, it's more a lib, that provides helper methods to parse and validate data. _About libs_: libs and utils are probably the same things, in that they provides a set of useful helpers, but a lib is more high-level code, when utils are just a collection of unrelated method about a same theme. To avoid unwanted complexity, say that every sets of helpers are a lib :smile:.
- Modal are handled by a query param (not active here, but will be enabled in a dedicated PR about routing)

### DIFF
--- a/client/app/components/accounts/wizard/creation.cjsx
+++ b/client/app/components/accounts/wizard/creation.cjsx
@@ -10,7 +10,7 @@ ReactDOM = require 'react-dom'
 Form    = require '../../basics/form'
 Servers = require '../servers'
 
-RequestsInFlightStore = require '../../../stores/requests_in_flight_store'
+RequestsStore = require '../../../stores/requests_store'
 
 RouterGetter = require '../../../getters/router'
 
@@ -32,12 +32,12 @@ module.exports = AccountWizardCreation = React.createClass
     displayName: 'AccountWizardCreation'
 
     mixins: [
-        StoreWatchMixin [RequestsInFlightStore]
+        StoreWatchMixin [RequestsStore]
         AccountMixin
     ]
 
 
-    # Build state form RequestsInFlightStore:
+    # Build state form RequestsStore:
     # - is an autodiscover request in action, or a provider available?
     getStateFromStores: ->
         state       = {}
--- a/client/app/components/accounts/wizard/creation.cjsx
+++ b/client/app/components/accounts/wizard/creation.cjsx
@@ -1,9 +1,10 @@
 {IMAP_OPTIONS} = require '../../../constants/defaults'
 {OAuthDomains} = require '../../../constants/app_constants'
 
-_        = require 'underscore'
-React    = require 'react'
-ReactDOM = require 'react-dom'
+_             = require 'underscore'
+React         = require 'react'
+ReactDOM      = require 'react-dom'
+AccountsUtils = require '../../../libs/accounts'
 
 Form    = require '../../basics/form'
 Servers = require '../servers'
@@ -16,7 +17,6 @@ AccountActionCreator = require '../../../actions/account_action_creator'
 RouterActionCreator = require '../../../actions/router_action_creator'
 
 StoreWatchMixin = require '../../../mixins/store_watch_mixin'
-AccountMixin    = require '../../../mixins/account_mixin'
 
 
 module.exports = AccountWizardCreation = React.createClass
@@ -25,7 +25,6 @@ module.exports = AccountWizardCreation = React.createClass
 
     mixins: [
         StoreWatchMixin [RequestsStore]
-        AccountMixin
     ]
 
 
@@ -41,7 +40,7 @@ module.exports = AccountWizardCreation = React.createClass
             OAuth:          RequestsGetter.isAccountOAuth()
 
         state.success = _.partial @redirect, account if account
-        _.extend state, @parseProviders discover if discover
+        _.extend state, AccountsUtils.parseProviders discover if discover
 
         return state
 
@@ -123,9 +122,11 @@ module.exports = AccountWizardCreation = React.createClass
 
         if @state.isDiscoverable and not(@state.imapServer or @state.smtpServer)
             [..., domain] = @state.login.split '@'
-            AccountActionCreator.discover domain, @sanitizeConfig @state
+            AccountActionCreator.discover domain,
+                AccountsUtils.sanitizeConfig @state
         else
-            AccountActionCreator.check value: @sanitizeConfig @state
+            AccountActionCreator.check
+                value: AccountsUtils.sanitizeConfig @state
 
 
     # Disable autodiscover when advanced settings are expanded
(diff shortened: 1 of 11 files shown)


m4dz:
Note: I have disabled some tests, so Travis may fail. I have a dedicated PR in WIP to fix all tests for the new components.

client/app/components/accounts/wizard/creation.cjsx
```diff
@@ -37,67 +29,21 @@ module.exports = AccountWizardCreation = React.createClass
...
-        _.defaults state,
-            isBusy:         isBusy
-            isDiscoverable: checkReq.status is null
-            alert:          null
-            success:        false
+        account  = RouterGetter.getAccountCreationSuccess()?.account
+        discover = RouterGetter.getAccountCreationDiscover()
+
+        state =
+            isBusy:         RouterGetter.getAccountCreationBusy()
```
misstick: Maybe `isAccountBusy()` should be better than `getAccountCreationBusy `.

client/app/components/accounts/wizard/creation.cjsx
```diff
@@ -37,67 +29,21 @@ module.exports = AccountWizardCreation = React.createClass
...
+            isDiscoverable: RouterGetter.getAccountIsDiscoverable()
```
misstick: Maybe `isAccountDiscoverable()` should be better than `getAccountIsDiscoverable `.

client/app/components/accounts/wizard/creation.cjsx
```diff
@@ -37,67 +29,21 @@ module.exports = AccountWizardCreation = React.createClass
...
+            alert:          RouterGetter.getAccountCreationAlert()
```
misstick: Maybe `isAccountAlert()` should be better than `getAccountCreationAlert `?

client/app/components/accounts/wizard/creation.cjsx
```diff
@@ -37,67 +29,21 @@ module.exports = AccountWizardCreation = React.createClass
...
+            OAuth:          RouterGetter.getAccountIsOAuth()
```
misstick: Maybe `isOauth()` should be better than `getAccountIsOAuth `?

client/app/components/accounts/wizard/creation.cjsx
m4dz: `isAccountCreationBusy()` (as what we wantr to known is "are the creation process spawned in background)?

client/app/components/accounts/wizard/creation.cjsx
m4dz: Hmmm, not sure: it returns the alert message, not just a boolean

client/app/getters/router.coffee
```diff
@@ -154,6 +155,75 @@ module.exports =
...
+        if discover.status is RequestStatus.SUCCESS
+            return true
+        # autodiscover failed w/o check in course: switch to manual config
+        else if discover.status is RequestStatus.ERROR and check.status is null
+            return false
+        else
+            return check.status is null
+
+
+    getAccountCreationAlert: ->
```
misstick: Obviously you can create a file `getters/request.coffee` to handle this.

client/app/getters/router.coffee
```diff
@@ -154,6 +155,75 @@ module.exports =
...
+        # autodiscover failed: set an alert only if check isn't already
+        # performed
+        else if discover.status is RequestStatus.ERROR and check.status is null
+            'DISCOVER_FAILED'
+
+        else
+            null
+
+
+    getAccountIsOAuth: ->
```
misstick: same as above.

client/app/getters/router.coffee
```diff
@@ -154,6 +155,75 @@ module.exports =
...
+        check = RequestsStore.get Requests.CHECK_ACCOUNT
+
+        if check.status is RequestStatus.ERROR
+            check.res.oauth
+        else
+            false
+
+
+    getAccountCreationDiscover: ->
```
misstick: same as above.

client/app/getters/router.coffee
```diff
@@ -154,6 +155,75 @@ module.exports =
...
+        discover = RequestsStore.get Requests.DISCOVER_ACCOUNT
+
+        if discover.status is RequestStatus.SUCCESS
+            discover.res
+        else
+            false
+
+
+    getAccountCreationSuccess: ->
```
misstick: same as above.

tests/client-units/account_store.coffee
```diff
@@ -65,15 +65,15 @@ describe 'AccountStore initialized without account', ->
...
-        error.should.equal RouterStore.getErrors().get('smtpPort')
+    # it 'Then RouterStore should have some errors', ->
+    #     should.exist RouterStore.getErrors().get('smtp')
+    #     error = RouterStore.getErrors().get('smtp')
+    #     error.message.should.equal 'translated config error smtp'
+    #     error.should.equal RouterStore.getErrors().get('smtpLogin')
+    #     error.should.equal RouterStore.getErrors().get('smtpPort')
 
 TEST_ACCOUNT =
     id: 'testid'
```
misstick: (reminder) We discussed @frankrousseau and I about this file, and this file is not in the right directory; it is functional tests not units.

misstick:
@m4dz about : 
````
The StoreWatchMixin in the Creation component: I assume a modal isn't directly relied to the application and won't inherit of all its states. Plus, I listen the RequestsStore rather than the RouterStore as long all its events loops comes from background requests only. I assume we improve this when we'll implement the real URL to modal (in query params) logic, as discussed
````

Did you add a `key`to `accountModalComponent` with specific information that would change and then update the component?
Maybe `StoreWatchMixin` is usefull in both views, but it should be more appropriate to also have a uniq key to `modalComponent`.

misstick:
@m4dz : about query params, I'm totally OK with you

m4dz:
@misstick ready to merge fopr me :)

m4dz:
AND BUILD PASSED :tada: :dancer:

PR #862, opened 2016-05-21
Add tests for Account, Contact, Layout and Message Stores

frankrousseau:
I put in this PR all tests related to stores. Every file contains a comment about should be fixed in the code, once the tests will allow to perform them safely.

### DIFF
--- a/client/test/account_store.spec.js
+++ b/client/test/account_store.spec.js
@@ -1,9 +1,54 @@
 'use strict';
 const assert = require('chai').assert;
+const _ = require('lodash');
 
 const mockeryUtils = require('./utils/mockery_utils');
 const SpecDispatcher = require('./utils/specs_dispatcher');
 const ActionTypes = require('../app/constants/app_constants').ActionTypes;
+
+const fixtures = {
+  account1: {
+    id: '123',
+    label: 'personal',
+    mailboxes: [
+      { id: 'a1', label: 'inbox', attribs: '', tree: [''], accountID: '123' },
+      { id: 'a2', label: 'sent', attribs: '', tree: [''], accountID: '123' },
+    ],
+  },
+  account2: {
+    id: '124',
+    label: 'pro',
+    mailboxes: [
+      {
+        id: 'b1', label: 'mailbox', attribs: '', tree: [''], accountID: '124',
+      },
+      {
+        id: 'b2', label: 'folder1', attribs: '', tree: [''], accountID: '124',
+      },
+    ],
+  },
+  account3: {
+    id: '125',
+  },
+  emptyAccount: {
+    label: '',
+    login: '',
+    password: '',
+    imapServer: '',
+    imapLogin: '',
+    smtpServer: '',
+    id: null,
+    smtpPort: 465,
+    smtpSSL: true,
+    smtpTLS: false,
+    smtpMethod: 'PLAIN',
+    imapPort: 993,
+    imapSSL: true,
+    imapTLS: false,
+    accountType: 'IMAP',
+    favoriteMailboxes: null,
+  },
+};
 
 
 describe('Account Store', () => {
@@ -11,9 +56,8 @@ describe('Account Store', () => {
   let dispatcher;
   let account1;
   let account2;
-  let account3;
 
-  function addAcount(account) {
+  function addAccount(account) {
     dispatcher.dispatch({
       type: ActionTypes.ADD_ACCOUNT_SUCCESS,
       value: { account },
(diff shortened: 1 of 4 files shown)


client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
...
+    account2.mailboxes = [
+      {
+        id: 'b1', label: 'mailbox', attribs: '', tree: [''], accountID: '124',
+      },
+      {
+        id: 'b2', label: 'folder1', attribs: '', tree: [''], accountID: '124',
+      },
+    ];
+    account3.id = '125';
+    addAcount(account1);
```
m4dz: I'm not sure all data-related stuff should belongs to the `before` statement. Maybe we should prefer fixtures, and make them loaded when we instanciate the accountStore by passing a `window.accounts` a9or something similar) that contains fixtures, rather than manually declares them.

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
...
+   */
+  describe('Actions', () => {
+    it('ADD_ACCOUNT_SUCCESS', () => {
+      dispatcher.dispatch({
+        type: ActionTypes.ADD_ACCOUNT_SUCCESS,
+        value: { account: { id: '125' } },
+      });
+      const accounts = accountStore.getAll();
+      assert.equal(accounts.get('125').get('id'), '125');
+      assert.deepEqual(accounts.get('125').get('mailboxes').toObject(), {});
```
m4dz: To complement the previous point, using fixtures will permit a more clever way to deep compare contents,

m4dz:
Thanks @frankrousseau for having take the time to make this one! I find you approach to split tests into _Actions_ and _Methods_ a clever way to cover properly Stores behaviors.

I have still a doubt about manipulating datas, and I ask myself about the opportunity to use external fixtures files to hydrate and tests stores. What do think about it? @misstick, any advice?

frankrousseau:
The upside of what you propose is that it makes things cleaner. But the downside is that tests data are splitted in another file and it makes tests harder to follow. My experience of testing is that it should require close to no thinking to be read. What do you think about it?

m4dz:
Hmmm, both ways have their pros and cons. Personnally, I prefer to let things separated 'coz we can have the need of reuse same fixtures accross some tests, but your solution have the advantage of having all the stuff we need in a straigth way.

I invoke @misstick to have its opinion :smile:

misstick:
@frankrousseau, @m4dz : I have no opinion about this; I have just seen that fixtures are not used even for server side tests: is this on purpose? I don't know.

We just have to be consistent with other cozy application: how is this handled elsewhere?

misstick:
@frankrousseau : At first, thank you for having starting that usefull work :) I give you my first impression about this PR.

Have you seen that other client unit tests exists, but they are not used anymore: `tests/client-units/account_store`. They seems to me very relevant, and I think you should take a look.

To my mind, this scope of this PR is too large; there is a lot of things to do for each store: enought to make a PR for each. In example, adding tests is not only write new code, but also removing old un-used code that has to do the job before.

frankrousseau:
I can't find the tests you're talking about. Are you talking about the Casper tests? These one are functional tests, they are not unit tests. They cover a broader scope but they are slow to run and unstable. They are different kind of tests, we should not compare them to unit tests.

About removing / cleaning the code, we need tests first. Once the unit tests will be written, it will be much easier to remove or rewrite store code because you will immediately noticed if you introduced regressions or if you broke something.

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
+'use strict';
+const assert = require('chai').assert;
+
+const mockeryUtils = require('./utils/mockery_utils');
```
misstick: I dont see and files with this path; did you forget to add it?

client/test/layout_store.spec.js
```diff
@@ -1,40 +1,41 @@
...
-    after(function () {
-        mockery.disable()
-        sandbox.restore()
-    })
-
-})
+'use strict';
+const assert = require('chai').assert;
+
+const mockeryUtils = require('./utils/mockery_utils');
```
misstick: same as before (no files found)

client/test/message_store.spec.js
```diff
@@ -0,0 +1,319 @@
+'use strict';
+const assert = require('chai').assert;
+
+const mockeryUtils = require('./utils/mockery_utils');
```
misstick: same as before (no files found)

client/test/contact_store.spec.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
+const assert = require('chai').assert;
+
+const mockeryUtils = require('./utils/mockery_utils');
```
misstick: same as before (no files found)

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
...
+const SpecDispatcher = require('./utils/specs_dispatcher');
```
misstick: same as before (no file found)

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
+'use strict';
+const assert = require('chai').assert;
```
misstick: I note that you used `ES6` syntax <3

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
...
+
+
+describe('Account Store', () => {
+  let accountStore;
+  let dispatcher;
+  let account1;
+  let account2;
+  let account3;
+
+  function addAcount(account) {
```
misstick: typo; should be `addAccount` instead.

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
...
+    addAcount(account2);
+  });
+
+
+  /*
+   * Problem noticed:
+   *
+   * * Remove mailbox event doesn't do anything.
+   * * getDefault can be used as getByMailbox.
+   */
```
misstick: Should you use `FIXME` and/or `TODO` comment previx instead of `Problem noticed`?

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
+'use strict';
+const assert = require('chai').assert;
+
+const mockeryUtils = require('./utils/mockery_utils');
+const SpecDispatcher = require('./utils/specs_dispatcher');
+const ActionTypes = require('../app/constants/app_constants').ActionTypes;
+
+
+describe('Account Store', () => {
+  let accountStore;
```
misstick: You should use 4 spaces for tabulation not 2.

misstick:
@frankrousseau : about previous tests, the absolute pat is: `./client/tests/client-units/account_store`.

client/test/account_store.spec.js
frankrousseau: With JS norm is two spaces.

misstick:
@frankrousseau : yes, I understand that. But I'm not talking about removing/cleansing dead code into application but about into tests. Look at my previous comment above.

frankrousseau:
I can't find them: https://github.com/cozy/cozy-emails/tree/master/client/tests

misstick:
@frankrousseau: My bad, the path was falsy :/ Try this: https://github.com/cozy/cozy-emails/tree/development/tests/client-units

frankrousseau:
Again I think they are a different kind of tests. These tests are functionnal too based on a scenario. What I did are very local tests which can be called unit tests. They are complementary' we should not oppose them.

client/test/account_store.spec.js
frankrousseau: ok thx

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
...
+    });
+    it('getByMailbox', () => {
+      const account = accountStore.getByMailbox('b1');
+      assert.equal(account.get('id'), '124');
+    });
+    it('getDefault', () => {
+      const account = accountStore.getDefault();
+      assert.equal(account.get('id'), '123');
+    });
+    it('getByLabel', () => {
```
misstick: @frankrousseau this method isn't used, you can remove this test and remove method from Store (@m4dz: can you confirm?)

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
...
+    it.skip('_setMailboxToImmutable', () => {});
+    it.skip('_initialize', () => {});
+    it.skip('_getByMailbox', () => {});
+    it.skip('_updateMailbox', () => {});
+    it.skip('_updateAccount', () => {});
+    it('getByID', () => {
+      const account = accountStore.getByID('124');
+      assert.equal(account.get('id'), '124');
+    });
+    it('getByMailbox', () => {
```
misstick: @frankrousseau this method isn't used anymore, you can remove this test and remove method from Store (@m4dz: can you confirm?)

client/test/account_store.spec.js
frankrousseau: Now that my PR about the test environment is included, I will fix this.

misstick:
@frankrousseau: Ok. But at first we wanted to clean files, and these functionnal tests are into `tests/client-units` directory; this doesnt make sense, dont you think so?

frankrousseau:
Yes, the folder name is misleading. I will rename it in another PR dedicated to improvements.

misstick:
@frankrousseau, @m4dz: about fixtures, here are fixtures `/tests/fixtures` that dont seem to be used; I can see into server side test includes `cozy-fixtures` but its not the same thing, isnt it?
Should we remove these files to remove ambiguity if we don't use fixtures at all?

frankrousseau:
Fixtures for server are different. They aimed to be stored in database and used as is. Here we talk about local memory fixtures. I'm not sure it's a good idea to try to mix them, because use cases are different and they are designed to work with the cozy-fixtures tool. I will take a look at it.

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
...
+
+
+  /*
+   * Problem noticed:
+   *
+   * * Remove mailbox event doesn't do anything.
+   * * getDefault can be used as getByMailbox.
+   */
+  describe('Actions', () => {
+    it('ADD_ACCOUNT_SUCCESS', () => {
```
misstick: @frankrousseau: You only test action `ADD_ACCOUNT_SUCCESS` but not `ADD_ACCOUNT_FAILURE`. You should test that no account is saved after a failure; you may also test that after dispatching several time the same `ADD_ACCOUNT_SUCCESS` action, only one account is created.

client/test/account_store.spec.js
```diff
@@ -0,0 +1,180 @@
...
+      const accounts = accountStore.getAll();
+      const mailbox = accounts.get('124').get('mailboxes').get('b1');
+      assert.equal(mailbox.get('label'), 'mailbox-edited');
+    });
+  });
+
+  describe('Methods', () => {
+    it.skip('_getMailboxIndex', () => {});
+    it.skip('_setMailboxToImmutable', () => {});
+    it.skip('_initialize', () => {});
```
misstick: I think you should test default accounts here

client/test/contact_store.spec.js
```diff
@@ -0,0 +1,106 @@
...
+          contacts.get('test2@right.com').get('address'), 'test@right.com');
+    });
+    it('CONTACT_LOCAL_SEARCH', () => {
+      let results = search('test');
+      assert.isDefined(results);
+      assert.equal(results.size, 3);
+      assert.isDefined(results.get('test@right.com'));
+      assert.isDefined(results.get('test2@right.com'));
+      assert.isDefined(results.get('test@cozy.io'));
+      results = search('Doe');
```
misstick: Don't forget to add empty lines between blocs

client/test/contact_store.spec.js
```diff
@@ -0,0 +1,106 @@
...
+        value: [contact1, contact2, contact3],
+      });
+      const contacts = contactStore.getAll();
+      assert.deepEqual(contacts.get('test@right.com').get('id'), contact1.id);
+      assert.deepEqual(contacts.get('test2@right.com').get('id'), contact1.id);
+      assert.deepEqual(contacts.get('john@cozy.io').get('id'), contact2.id);
+      assert.equal(
+          contacts.get('test2@right.com').get('address'), 'test@right.com');
+    });
+    it('CONTACT_LOCAL_SEARCH', () => {
```
misstick: Don't forget guide-lines https://github.com/cozy/cozy-guidelines#generic-code-style

client/test/contact_store.spec.js
```diff
@@ -0,0 +1,106 @@
...
+    });
+  });
+
+  describe('Methods', () => {
+    it('getByAddress', () => {
+      assert.deepEqual(contactStore.getByAddress('test@right.com').toObject(),
+                       contact1);
+      assert.isUndefined(contactStore.getByAddress('test@wrong.com'));
+    });
+    it('getAvatar', () => {
```
misstick: Don't forget to add comments to explain why where expecting this instead of that.

client/test/message_store.spec.js
```diff
@@ -0,0 +1,319 @@
...
+
+  before(() => {
+    dispatcher = new SpecDispatcher();
+    mockeryUtils.initDispatcher(dispatcher);
+    mockeryUtils.initForStores();
+    messageStore = require('../app/stores/message_store');
+  });
+
+  /*
+   * Problems noticed:
```
misstick: Should you use `FIXME` and/or `TODO` comments instead?

client/test/message_store.spec.js
```diff
@@ -0,0 +1,319 @@
...
+      };
+      dispatcher.dispatch({
+        type: ActionTypes.MESSAGE_FLAGS_SUCCESS,
+        value: { updated: [message1, message2] },
+      });
+      const messages = messageStore.getAll();
+      assert.deepEqual(messages.get('ai1').toObject(), message1);
+      assert.deepEqual(messages.get('ai2').toObject(), message2);
+    });
+    it.skip('MESSAGE_MOVE_SUCCESS', () => {
```
misstick: Why do you skip this?

client/test/message_store.spec.js
```diff
@@ -0,0 +1,319 @@
...
+    });
+    it.skip('MESSAGE_SEND_SUCCESS', () => {
```
misstick: Why do you skip this? You should write a comment if its not an oversight ;)

misstick:
@frankrousseau: (global comment) I noticed that this PR is adding unit test for some Stores but not all of them; why? Can you specify it into PR description please?

I usually add checklist because I find them more clear than big sentences as you please :)

client/test/account_store.spec.js
m4dz: I never use it too. I assume we can remove it safely.

client/test/account_store.spec.js
m4dz: Like the previous one, but I suggest I'll handle them on my next PR about dead code cleaning :)

m4dz:
> I'm not sure it's a good idea to try to mix them, because use cases are different

I totally agree, fixtures sets are dedicated to their parts (server vs client) and we shouldn't mix them.

client/test/account_store.spec.js
misstick: @frankrousseau : I'm OK to test this in an other PR

client/test/account_store.spec.js
misstick: @frankrousseau: I assume that these comments are about improvement. Could you move this into the right trello board (ie. with a card called "Store:improvements") so that can be handled asap please?
Same comments for other comments about improvements.

misstick:
@frankrousseau : I'm waiting about your fixes so that I can merge this PR :)

frankrousseau:
As @misstick and @m4dz  requested, I did some improvements:

* Fix mockery utils usage;
* Put fixtures on top of the test file for each test file;
* Remove useless skipped tests of private methods;
* Better formation of problems found in store code while writing tests;
* Fix addAccount typo in the function name.

I you agree with theses fixes, we should merge this PR and tackle other problems, like the *unit-test* folder renaming, in another PR.

misstick:
@frankrousseau : I will review it tomorrow. Thanks a lot! :)

client/test/contact_store.spec.js
misstick: @frankrousseau: I still not find this file; maybe you have forgotten to re-add it when you rebase your commits?

frankrousseau:
@misstick I didn't see any rules related to prefixing or commiting. So, I didn't guess that `[review]` should be used. 

Sorry, for the melted commits but my context was not right to do it (I had separated branch for tests configuration changes and tests related to accounts). Another thing is that writing tests on code I didn't write is super boring, so making it perfect is even more painful. Now you know that, I hope you will understand my tradeoffs about quality.

PR #864, opened 2016-05-24
Improve account creation wizard layout

m4dz:
This PR improve account creation wizard layout and introduces some behaviors (like autofocus, or credential fields auto-fading)


client/app/components/accounts/server.cjsx
```diff
@@ -118,3 +111,13 @@ module.exports = AccountServer = React.createClass
         _.each state, (value, key) ->
             globalState["#{protocol}#{key[0].toUpperCase()}#{key.slice 1}"] = value
         @props.onChange globalState
+
+
+    _securityOptions: ->
```
misstick: Why are you using `_`as a prefix? I think that if you really want this method to be private, you should remove it from ReactComponent or do nothing else than removing the prefix :)

client/app/components/accounts/server.cjsx
```diff
@@ -56,15 +56,8 @@ module.exports = AccountServer = React.createClass
...
-            .map (protocol) ->
-                value: protocol
-                label: t "server protocol #{protocol}"
-            .concat
-                value: 'none'
-                label: 'server protocol none'
-
-        <div className="content">
+        <div className="content server">
+            <h2>{@props._protocol}</h2>
```
misstick: Why are you using `_`as a prefix? I think that if you really want this method to be private, you should remove it from ReactComponent or do nothing else than removing the prefix :)

misstick:
👍 I'm waiting for other merge/rebase to merge this one

m4dz:
@misstick ready to merge!

m4dz:
@misstick rebase OK :+1:

PR #866, opened 2016-05-26
Creation wizard lib tests

m4dz:
Tests for the Account creation wizard lib:

- [x] validateState
- [x] sanitizeConfig
- [x] parseProviders

### DIFF
--- a/client/test/accounts_lib.spec.js
+++ b/client/test/accounts_lib.spec.js
@@ -92,4 +92,90 @@ describe("Accounts libs spec", () => {
 
     })
 
+
+    describe("Parse providers to return part of state", () => {
+
+        let providers, nextState
+
+        beforeEach(() => {
+            nextState = AccountsLib.parseProviders(providers)
+        })
+
+        describe("Parse valid provider settings", () => {
+            before(() => {
+                providers = [{
+                    "type": "imap",
+                    "hostname": "mail.gandi.net",
+                    "port": "993",
+                    "socketType": "SSL"
+                }, {
+                    "type": "imap",
+                    "hostname": "mail.gandi.net",
+                    "port": "143",
+                    "socketType": "STARTTLS"
+                }, {
+                    "type": "pop3",
+                    "hostname": "mail.gandi.net",
+                    "port": "995",
+                    "socketType": "SSL"
+                }, {
+                    "type": "pop3",
+                    "hostname": "mail.gandi.net",
+                    "port": "110",
+                    "socketType": "STARTTLS"
+                }, {
+                    "type": "smtp",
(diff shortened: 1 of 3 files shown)


client/test/accounts_lib.spec.js
```diff
@@ -0,0 +1,95 @@
...
+const AccountsLib = require('../app/libs/accounts')
+
+
+describe("Accounts libs spec", () => {
+
+    describe("Validate and returns a new state", () => {
+
+        it("should returns a simple value in a new state", () => {
+            let state = {}
+            let nextState = AccountsLib.validateState('login', 'mail@cozy.io', state)
```
jsilvestre: Out of curiosity: why do you use let instead of const here?

m4dz: Coz' in some tests I need to change their values (e.g. https://github.com/cozy/cozy-emails/pull/866/files/769ab196b3f04156aa8bbc009ccf4ce34e6f3779#diff-75f7b93415035b03c811d0173ae2f7fbR58) so I try to stay consistent and use `let` everywhere. But I'm open to another solution if you provide one :smile:

jsilvestre: That's what I thought, I was not sure I would have kept with consistency in this case. I wondered if there was something else. Thanks!

client/test/accounts_lib.spec.js
```diff
@@ -0,0 +1,95 @@
...
+            assert.propertyVal(nextState, 'imapLogin', nextState.login)
+            assert.propertyVal(nextState, 'smtpLogin', nextState.login)
+
+            nextState = AccountsLib.validateState('login', 'box@cozy.io', nextState)
+
+            assert.propertyVal(nextState, 'imapLogin', nextState.login)
+            assert.propertyVal(nextState, 'smtpLogin', nextState.login)
+        })
+
+        it("should leave imap / smtp login properties untouched when they differ from login", () => {
```
misstick: @m4dz : shouldnt you use a new group such as: `imap / smtp properties` for more lisibility?

misstick:
@m4dz : I don't see the point of prefixing commit with `[test]` because its the goal of this PR; I think that you can a flag to the PR, such as `test`. This is done form `enhancement` or `bug`, why not for tests?

Look at my previous comment to add more lisibility to your test; otherwhise everything looks good to me! 👍

m4dz:
>  I don't see the point of prefixing commit with [test]

Because of https://github.com/commitizen/cz-cli#using-the-command-line-tool :smiley:. We don't use explicitely Commitizen, but I use their prefix convention for commits.

PR #868, opened 2016-05-27
[fix] Gmail mailboxes issues 

misstick:
- [x] Inbox belongs to `flaggedMailbox`; it should appear on top of mailboxes list,
- [x] Inbox isnt found for gmail accounts,
- [x] spinner is always visible into `[Gmail]` mailbox,
- [x] special mailbox icons are not found (`drafts`, `deletedMessages`, etc.).

### DIFF
--- a/client/app/components/application.coffee
+++ b/client/app/components/application.coffee
@@ -99,7 +99,6 @@ module.exports = React.createClass
                     accounts        : RouterGetter.getAccounts()?.toArray()
                     composeURL      : composeURL
                     newAccountURL   : newAccountURL
-                    mailboxes       : RouterGetter.getMailboxes()
                     nbUnread        : RouterGetter.getUnreadLength()
                     nbFlagged       : RouterGetter.getFlaggedLength()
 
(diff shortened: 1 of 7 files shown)


misstick:
@aenario: I found that gmailInbox isnot flagged as it should; `mailbox.tree` is not the same as OVH, ie. We have to look at `mailbox.attribs` instead. Did you know it? I'd love you tell me that long story :)

misstick:
@m4dz : can you review this PR; its quite old and finished :(

frankrousseau:
ping @aenario Can you help on some emails review?

m4dz:
I'm on it (sorry for the delay :/)

aenario:
@misstick If i recall correctly, the idea is that we should trust mailbox.attribs when possible.

However, 
- some providers dont have attribs to their mailboxes at all, so we attempt to guess by the name.
- in some cases (like every cozy team member who switched from gmail to OVH), the hierarchy is weird (INBOX > INBOX / INBOX > Sent), see tests fixtures. This was an attempt to have the displayed tree cleaner in these cases.

m4dz:
Thanks :tada: ! (and again, sorry for the delay)

PR #875, opened 2016-06-19
Add spinner button on creation

m4dz:
Add a spinner on submit button in create account modal when process is busy (discover, check…)


frankrousseau:
Thx @m4dz !

PR #876, opened 2016-06-21
Fix gateway timeouts

m4dz:
This PR tries to fix gateway timeouts by throttling huge / heavy requests.

*Context*: sometimes, the server node process is struggled by a too high number of requests (especially w/ Node < 4). This may result to _Gateway Timeouts_ requests as long as the server is still idle w/ previous concurrent requests. So we add a throttle (a plugin for superagent) that will delay requests from client to the server part to minimize the load and keep server below the struggle point. It applies on all `GET` requests, and on some heavy `PUT`/`POST` requests (like batch updates) that may overload the server.

### DIFF
--- a/client/app/components/menu.coffee
+++ b/client/app/components/menu.coffee
@@ -53,7 +53,6 @@ module.exports = Menu = React.createClass
                                 div
                                     className: 'account-details',
                                         span {}, @props?.search
-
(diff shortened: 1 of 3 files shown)


misstick:
@m4dz : could you add the context and use cas when it appears plz ?

PR #874, opened 2016-06-19
[test] fix broken client tests

m4dz:
This PR include some client tests fixes, and (finally) enable client tests runs on CI. It also prepares `RequestsStore` spec.

PR #873, opened 2016-06-10
Add remove message feature

misstick:
### When a message is deleted from `messageComponent`:
 - if conversation still exist (conversation has other messages): 
  - if next message exist: select it,
  - if not select previous message,
  - in both cases: update `:messageID` into`URL`.
 - if not, then conversation is deleted and:
  - `messageListComponent` is updated,
  - select next conversation (upper in the list),
  - update `:conversation:messageID` into URL,

### How is defined nearest message?
 - get next message from conversation,
 - otherwise get previous message from conversation,
 - if message was the last of the conversation: select next conversation into `messageListComponent`,
 - if no nearest message was found then close `conversationComponent`.
 - if page is refreshed by browser by external behaviour (such as browser refresh button or websocket event), nearest message should be found too.

### What is behind `messagesListCounter`?
  - it is the length messages into conversation which takes into account `currentFilter`.
  - if a conversation is displayed into `unseenMailbox`, the counter is only about unread messages from the conversation.

### Deleting a message from `unreadMailbox`
 - [x]  `unreadCounter` should be updated into `mailboxMenu`,
 - if conversation have still unread messages:
  - [x]  update `unreadCounter` into `messageListItem` of mailbox,
  - [x]  remove message from `conversationComponent`,
  - [x]  select the `nearestMessage`,
  - [x]  update URL with `nearestMessageID`.
 - otherwhise:
  - [x] remove message from `messageList`,
  - [x]  select the `nearestMessage`,
  - [x]  update URL with `nearestConversationID/IDnearestMessageID`.
 

### Deleting a message from `trashBox` cant happen
  - [x] `removeButton` must be hidden in this case.

## tasks
 - [x] add `toolbarMessageComponent` with the only button `deleteMessage`,
 - [x] this toolbox only appear for `noTrash` mailbox,
 - make `deleteMessageFeature` work:
  - [x] delete message into `messageStore`,
  - [x] update message list into `routerStore`,
  - [x] remove `messageItem` from `messageListComponent`,
  - [x] remove `messageItem` from `conversationComponent`,
  - if `conversationComponent` is empty:
    - [x] remove conversation from `messageStore`,
    - [x] remove conversation from `messageListComponent`
    - [x] selected nearest conversation,
    - [x] otherwise close conversation,
    - [x] update counter into concerned components: `messageLisItemComponent` or/and `mailboxItemCounterComponent`.

## issues
 - [x] `getMoreMessages` button didnt work on `trashbox`,
[...]

PR #870, opened 2016-06-03
[tests] Add tests for router store

frankrousseau:
Here are unit tests for the route store.

I skipped three tests:

* ACCOUNT_CREATION_SUCCESS action test because it requires 5s to be tested properly
* gotoNextConversation() and gotoPreviousConversation(), I don't understand how they work and what they do.

PR #878, opened 2016-06-22
[test] fix and clean basic tests to add complex one (such as remove message)

misstick:
wait for these merge before merge this one:
- [fix] tests https://github.com/cozy/cozy-emails/pull/874,
- [test] routerStore https://github.com/cozy/cozy-emails/pull/870,
- [feature] remove message https://github.com/cozy/cozy-emails/pull/873,
- [config] js files https://github.com/cozy/cozy-emails/pull/879

### TODO
 - [x] add improvement asked here: https://github.com/cozy/cozy-emails/pull/870
 - [x] fix `MessageStoreTests`,
 - [x] fix `AccountStoreTests`,
 - [ ] fix `ContactStoreTests`,
 - [ ] fix `RouterStoreTests`,
 - [x] fix consistency issues: add fixture modules instead of using a local objet,
 - [x] fix test workflow: reset context after each.

### DIFF
--- a/client/app/stores/message_store.coffee
+++ b/client/app/stores/message_store.coffee
@@ -5,6 +5,14 @@ Store = require '../libs/flux/store/store'
 
 {ActionTypes, MessageFlags, MessageFilter} = require '../constants/app_constants'
 
+
+# FIXME: server side data
+#  - The fact that the conversation length is calculated remotely,
+#  is not clear for the developer.
+#
+#  - Most operations, like getting raw messages,
+#  don't update conversation length.
+
 class MessageStore extends Store
 
     ###
@@ -17,7 +25,6 @@ class MessageStore extends Store
 
 
     _updateMessages = (result={}, timestamp) ->
-
         {messages, conversationLength} = result
 
         # This prevent to override local updates
@@ -49,7 +56,6 @@ class MessageStore extends Store
         message.mailboxID = mailboxIDs.shift()
 
         if _shouldUpdateMessage message, timestamp
-
             attachments = message.attachments or Immutable.List []
 
             message.date          ?= new Date().toISOString()
@@ -69,7 +75,6 @@ class MessageStore extends Store
                 return """
                     #{message.id} "#{message.from[0].name}" "#{message.subject}"
                 """
-
             _messages = _messages.set message.id, messageMap
 
 
--- a/client/app/stores/message_store.coffee
+++ b/client/app/stores/message_store.coffee
@@ -168,9 +168,9 @@ class MessageStore extends Store
 
         handle ActionTypes.MAILBOX_EXPUNGE, (mailboxID) ->
             _messages = _messages.filter (message) ->
-                test0 = mailboxID isnt message.get('mailboxID')
-                test1 = not (mailboxID of message.get 'mailboxIDs')
-                test0 or test1
+                isSameMailbox = mailboxID is message.get('mailboxID')
+                hasSameMailbox = mailboxID of message.get 'mailboxIDs'
+                not isSameMailbox or not hasSameMailbox
             .toOrderedMap()
             @emit 'change'
 
--- a/client/app/stores/message_store.coffee
+++ b/client/app/stores/message_store.coffee
@@ -45,11 +45,15 @@ class MessageStore extends Store
 
     _saveMessage = (message) ->
         message = _.cloneDeep message
-
-        # Save reference mailbox into message informations
-        message.mailboxID = _.keys(message.mailboxIDs).shift()
+        oldMessage = _messages.get(message.id);
 
         if _shouldUpdateMessage message
+
+            # Save reference mailbox into message informations
+            if not _.isString(message.mailboxID) or _.isEmpty(message.mailboxID)
+                message.mailboxID  = oldMessage?.get('mailboxID')
+                message.mailboxID ?= _.keys(message.mailboxIDs).shift()
+
             message.date          ?= new Date().toISOString()
             message.createdAt     ?= message.date
             message.flags         ?= []
--- a/client/app/stores/message_store.coffee
+++ b/client/app/stores/message_store.coffee
@@ -194,14 +194,18 @@ class MessageStore extends Store
 
 
     isUnread: ({flags=[], message}) ->
+        if message and message not instanceof Immutable.Map
+            message = Immutable.Map message
         if message?
             flags = message.get('flags') or []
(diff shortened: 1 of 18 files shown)


frankrousseau:
Can you make sure that the linter returns no error?

m4dz:
@frankrousseau linters are a complex problem IMHO: never same linters nor version, sparse configs… If I can, I suggest we let @misstick finished the objective of this PR, and I'll open another one to address specific linters issues / config :)

frankrousseau:
OK to finish this PR without linting. But the linter is already configured. So, there is not much work to do about it.

client/test/router_store.spec.js
```diff
@@ -188,401 +173,500 @@ describe('Router Store', () => {
...
-          lastPage: {
-            info: 'last-page',
-            isComplete: true,
-          },
-          result: {},
-          timestamp: new Date(),
-        },
+
+
+      it('Select Tab from AccountEdit page', () => {
```
frankrousseau: We have a question here about whether these tests should be considered as unit tests or functional tests. The ROUTE_CHANGE event looks tricky to test and requires more functional testing than other events. 
But, it raises another question, should testing events be considered as unit tests of functional tests?

Whatever the answer to that questions do not have any consequence on what you are doing. I mean, we should discuss later of the file organization of the email tests.

misstick:
@m4dz : I've not finished the 1rst scope, but:
 - RouterStore will tool some days (maybe 3-4),
 - ContactStore are small test but i would imply:
   - adding tests on `messagesStore` to handle `to`, `cc`, `from` etc. fields,
   - add `fixtures/contact.js` to get these data from several spec files,
   - then clena contact tests.
 - and I think that requestsStore should be improved so that it will take few time to add tests

aenario:
If possible, avoid using new Date() in tests as it is not pure and cause intermittence in my tests.

client/app/stores/message_store.coffee
```diff
@@ -147,7 +168,9 @@ class MessageStore extends Store
 
         handle ActionTypes.MAILBOX_EXPUNGE, (mailboxID) ->
             _messages = _messages.filter (message) ->
-                not (mailboxID of message.get 'mailboxIDs')
+                test0 = mailboxID isnt message.get('mailboxID')
+                test1 = not (mailboxID of message.get 'mailboxIDs')
+                test0 or test1
```
m4dz: Namming is hard, isn't it :wink: ?

client/app/stores/router_store.coffee
```diff
@@ -364,6 +373,8 @@ class RouterStore extends Store
             # Message should be in mailbox
             inMailbox = mailboxID of message.get 'mailboxIDs'
 
+            # console.log 'PLOP', message.get 'mailboxIDs'
```
m4dz: Oups…

client/app/stores/router_store.coffee
```diff
@@ -437,6 +448,10 @@ class RouterStore extends Store
         _URI
 
 
+    getMessagesPerPage: ->
+        MSGBYPAGE
```
m4dz: Not sure what this one means…

misstick: What are you talking about: const name or public method name?

client/test/contact_store.spec.js
```diff
@@ -1,123 +1,123 @@
...
-      assert.isFalse(contactStore.isExist('test@wrong.com'));
-      assert.isTrue(contactStore.isExist(email1a));
-    });
-  });
-
-  after(() => {
-    mockeryUtils.clean();
-  });
-});
+// 'use strict';
```
m4dz: Why not removing this file, or adding `skip` ti bypass tests, instead of commenting everything?

client/test/layout_store.spec.js
```diff
@@ -1,41 +1,41 @@
...
-      });
-      assert.equal(layoutStore.isIntentAvailable(), true);
-    });
-  });
-
-  after(() => {
-    mockeryUtils.clean();
-  });
-});
+// 'use strict';
```
m4dz: Like contacts store spec…

client/test/requests_store.spec.js
```diff
@@ -1,20 +1,20 @@
...
-  let dispatcher;
-
-  before(() => {
-    dispatcher = new SpecDispatcher();
-    mockeryUtils.initDispatcher(dispatcher);
-    mockeryUtils.initForStores(['../app/stores/requests_store']);
-    requestsStore = require('../app/stores/message_store');
-  });
-})
+// 'use strict'
```
m4dz: Like contacts store spec

client/test/router_store.spec.js
```diff
@@ -1,590 +1,505 @@
...
-      routerStore.gotoPreviousMessage();
-      routerStore.gotoPreviousMessage();
-      assert.equal(routerStore.gotoNextMessage().get('me3'));
-    });
-  });
-  after(() => {
-    mockeryUtils.clean();
-  });
-});
+// 'use strict';
```
m4dz: Like contacts store spec

client/test/utils/specs_dispatcher.js
```diff
@@ -6,14 +6,12 @@ TestDispatcher.prototype.register = function register(callback) {
...
-
 TestDispatcher.prototype.dispatch = function dispatch(payload) {
   this._callbacks.forEach((callback) => {
     callback.call(this, payload);
   });
 };
 
-TestDispatcher.prototype.waitFor = function waitFor(events) {
-};
+TestDispatcher.prototype.waitFor = function waitFor() {};
```
m4dz: Shouldn't we add a real version of the `waitFor` method instead of fakingly mocking it? Does tests pas correctly even?

misstick: In fact in tests cases it is juste useless; needn't so much complexity.

client/app/stores/router_store.coffee
m4dz: Agree, but what does it do lonely here ?

client/test/utils/specs_dispatcher.js
m4dz: al'right :+1:

client/app/stores/router_store.coffee
misstick: It will be usefull for `router_store.specs` but you're right, it is no used, but should be... I remove it?

PR #882, opened 2016-06-24
[test] Remove legacy test folder

frankrousseau:
The `client/tests` folder is not used anymore. It's better to remove it to avoid any confusion.


misstick:
@frankrousseau: is it finished?

m4dz:
Right! Thanks @frankrousseau :tada:

PR #883, opened 2016-06-27
[lint] Add server code linting to Continuous Integration

frankrousseau:
I fixed linting (I only removed errors, warning are still present) and made linting available on Travis continuous integration.


frankrousseau:
Please do not review / merge yet, some code is not linted.

frankrousseau:
OK, this PR is mergeable.

frankrousseau:
Oups, no... travis is red...

package.json
```diff
@@ -43,6 +43,7 @@
   "devDependencies": {
     "chai": "3.5.0",
     "coffee-19-jshint": "2.3.1",
+    "coffee-jshint": "0.2.3",
```
m4dz: It seems that we have droped coffeescript 1.9 support, so I suggest we remove coffee-19-jshint to stuck to coffee-jshint official release.

package.json
```diff
@@ -78,8 +79,10 @@
     "fixtures:load": "cozy-fixtures load tests/fixtures",
     "fixtures:loader": "coffee tests/fixtures/loader.coffee",
     "lint": "npm-run-all 'lint:*'",
-    "lint:coffeelint": "coffeelint --file coffeelint.json --quiet ./server/**/*.coffee",
-    "lint:jshint": "coffee-jshint --options node ./server/**/*.coffee",
+    "lint:server": "npm-run-all 'lint:server:*:*'",
+    "lint:server:code:coffeelint": "./node_modules/.bin/coffeelint --file coffeelint.json  server.coffee ./server/*.coffee ./server/**/*.coffee",
```
m4dz: I'm not sure that we need to be too much precise with tasks namespace. What for a `lint:server:coffee` (which lint codebase and tests) and `lint:server:js`(for the js part)?

m4dz: You shouldn't explicitely use `./node_modules/.bin` prefix: npm scripts add it to the PATH env, so directly call `coffeelint` in script will make npm use primarily the locally installed package :wink:.

tests/client-units/account_store.coffee
```diff
@@ -44,8 +44,8 @@ describe 'AccountStore initialized without account', ->
     #     RouterStore.isWaiting().should.be.false
 
     # it 'Then AccountStore should have a nomailboxes error', ->
-    #     noMailboxErr = RouterStore.getErrors().get('nomailboxes')
-    #     noMailboxErr.message.should.equal 'translated config error nomailboxes'
+    #    noMailboxErr = RouterStore.getErrors().get('nomailboxes')
```
m4dz: I think the space here is ok (due to commenting, in fact). But souldn't we disable tests using `skip` instead of commenting them?

frankrousseau: What about removing them? They are here for long. I'm not sure it makes sense to keep commmented tests.

package.json
frankrousseau: Ok, but are you sure that there is no conflict with a global install?

package.json
frankrousseau: Ok but first, let's ask to @aenario why he chosed this package over the official one.

Whatever, we should only have one package.

m4dz: I agree. I think it was because we were stucked to coffeescript 1.9, but it isn't the case anymore. Let's wait for @aenario ;)

package.json
m4dz: Yup, npm script command first search in _node_modules_ to find executable, and then fallback to global env if it doesn't find it.

tests/client-units/account_store.coffee
m4dz: Right for me: if they're not relevant, go to suppress them instead of just skipping them :)

package.json
frankrousseau: @aenario told me that the 19 version could ve removed.

m4dz:
@frankrousseau nice :+1: !

What about scripts namespaces ? Isn't better to limit to 3 levels (`lint:server:coffee`) or are the specificity really relevant? /poke @aenario if you've got any thoughts?

frankrousseau:
The problem is that coffelint and coffejshint are very different. I think it's time to move forward: overall benefits of this PR overcome this namespace discussion. If you want to change namespaces, you can propose another PR and we will have the discussion there.

m4dz:
Thanks @frankrousseau :tada:

PR #895, opened 2016-07-12
[tests] Add ActionCreator 1rst tests

misstick:
##### 1. Add backbone for:
 - [x] `routerActionCreator`,
 - [x] `accountActionCreator`,
 - [x] `messageActionCreator`,

##### 2. Add tests for:
 - [x] `messageActionCreator.deleteMessage()` from a message.

##### 3. Spy for:
 - [x] success callback,
 - [x] error callback.

### DIFF
--- a/client/test/message_action_creator.spec.js
+++ b/client/test/message_action_creator.spec.js
@@ -93,7 +93,8 @@ describe('MessagesActionCreator', () => {
       let spySend;
       let callback;
       let action = 'mon action';
-      let message = { conversationID: 'plop', text: 'coucou' };
+      const message = { conversationID: 'plop', text: 'coucou' };
+
 
       beforeEach(() => {
         if (spySend === undefined) {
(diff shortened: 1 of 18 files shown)


client/test/message_action_creator.spec.js
```diff
@@ -0,0 +1,246 @@
...
+        // 1rst one for REQUEST
+        // Last one for SUCCESS
+        assert.equal(spyDispatcher.callCount, 2);
+
+        let args = spyDispatcher.getCall(1).args[0];
+        assert.equal(args.type, type);
+
+        let res = { conversationID: 'plop', text: 'coucou', html: 'coucou'};
+        assert.equal(args.value.action, action);
+        _.each(JSON.parse(args.value.message), (value, key) => {
```
m4dz: Shouldn't a [`deepEqual`](http://chaijs.com/api/assert/#method_deepequal) better in this case?

m4dz:
Great!

I ask myself about creating a SpecSuperagent instead of just mocking requests with Sinon?

misstick:
wouldnt it be less readable?

client/test/message_action_creator.spec.js
misstick: In this case, we cant to compare:
 `let res = { "conversationID": "plop", "text": "coucou", "html": "coucou"};`with `let res = { conversationID: 'plop', text: 'coucou', html: 'coucou'};` and `deepEqual` doesnt resolve this issuer

m4dz: Seen from here, those 2 are strictly equals and `deepEqual` shoudl do the trick, nope?

m4dz:
I hope not, and this is the gopal opf the mocking feature. But nwm, we can stuck at this one for this and we'll try with mocking later, al'right?

misstick:
I'm OK with it.

client/test/message_action_creator.spec.js
misstick: nope :( I tried...

PR #902, opened 2016-07-22
[DO NOT MERGE] Redux

aenario:


m4dz:
Woohoooo \o/ ! Nice shot!

Firstly, this is just an awesome POC, and I'm really (really really) happy to see it happen. And your proposal for this migration is really pretty, hope it mean the rest will be so easy too.

Here my though about:

1. I see you always rely on Getters, even in ActionCreators or in others Getters. Do we consider that the only good way to retrieve content form Redux Store is to use Getters everywhere (I'm totally fine with that, just a matter to confirm it)
2. I just :heart: so much using `props` everywhere in all components including `Application` and the _redux-connect_-like solution, seems perfect to me!
3. I see that you've isolated some curation logics in `utils`. We've previously try to avoid `utils` in favor of `libs`, just to try to not put too many things everywhere. We just need to sync our watches by documenting which part of code does what and where.

client/app/reducers/selection.coffee
```diff
@@ -0,0 +1,38 @@
+Immutable = require 'immutable'
+
+{ActionTypes} = require '../constants/app_constants'
+
+DEFAULTSTATE =
```
gregorylegarec: I think that we should use underscore separated words in all our constants (`DEFAULT_STATE` instead of `DEFAULTSTATE`), like for example `MAILBOX_SELECT_ALL` below.

client/app/reducers/selection.coffee
```diff
@@ -0,0 +1,38 @@
...
+    selected: Immutable.Set()
+    allSelected: false
+
+module.exports = (state = DEFAULTSTATE, action) ->
+    switch action.type
+        when ActionTypes.MAILBOX_SELECT_ALL
+            if state.allSelected
+                nextState = DEFAULTSTATE
```
gregorylegarec: What about having a dedicated action `ActionTypes.MAILBOX_DESELECT_ALL`. It seems that this part is too much related to a checkbox UI action. If we consider only this checkbox, it makes sense to select/unselect all on every check, i.e. dispatch a `SELECT_ALL` action. But maybe at

client/app/reducers/selection.coffee
```diff
@@ -0,0 +1,38 @@
...
+            else
+                nextState =
+                    allSelected: true,
+                    selected: Immutable.Set()
+
+        when ActionTypes.MAILBOX_SELECT
+            id = action.value.id
+            if(action.value.value and not state.selected.has(id))
+                nextState =
+                    allSelected: state.allSelected,
```
gregorylegarec: Don't `allSelected` should be false except when `id` was the only message not selected ?

client/app/reducers/selection.coffee
```diff
@@ -0,0 +1,38 @@
...
+    switch action.type
+        when ActionTypes.MAILBOX_SELECT_ALL
+            if state.allSelected
+                nextState = DEFAULTSTATE
+            else
+                nextState =
+                    allSelected: true,
+                    selected: Immutable.Set()
+
+        when ActionTypes.MAILBOX_SELECT
```
gregorylegarec: Same remark as for `MAILBOX_DESELECT_ALL`, what about a `MAILBOX_DESELECT` action ?

client/app/reducers/selection.coffee
aenario: agreed

client/app/reducers/selection.coffee
aenario: Until we move on to saga or another redux middleware, actions should be considered as events, ideally in the past tense ("SOMETHING_HAPPENED"). This action "full" name could be `USER_CLICKED_SELECT_ALL_BUTTON`, if we add more button, we can add more actions. 
if another situation require us to select all messages (for instance, going to next page with select all enable across pages), we should handle it in the appropriate action ("USER_CLICKED_NEXT_PAGE")

The other extreme is to have actions as setters (SET_CURRENT_MESSAGE), in which case, you end up calling several action in a row from a ActionCreator, you have to be sure you call them in the proper order at call site, you have to be sure you are calling all of them and so on.

client/app/reducers/selection.coffee
aenario: yes it should :-)

client/app/reducers/root.coffee
```diff
@@ -0,0 +1,10 @@
+
+selectionReducer = require './selection'
+messagesReducer = require './message'
+{combineReducers} = require('redux')
```
gregorylegarec: Not conform use of parenthesises
```coffeescript
require 'redux'`
```

PR #904, opened 2016-07-29
Migrate Contact Store to Contact Reducer using Redux

gregorylegarec:
Please @aenario have a review, I d'like to have your opinion about differents
points:
* If I used Immutable correctly
* If my redux reducer is OK

Aslo, I spent some time refactoring the part of the code in charge of mapping
raw contacts to immutable contacts and raw contacts list onto an immubtable map.
I choose to create a dedicated library in libs/mappers, tell me if it is ok for
you.

Also, I choose to change a little bit the logic to have :

* a mapper function which map a raw contact to an immutable, without dealing
with the way they're will be mapped in an immutable map. It means that I add
a property `addresses` which keep all the email address of a contact. I also
kept the `address` property which store the contact primary email (it is also
used to to store the email used as index).
* a mapper function which map a raw contact list to an array of immutables
* a mapper function which return a mutator, to pass to
`Immutable#Map#withMutations` method. I choose to keep the raw contacts to
immutables logic outside, so we have to pass an array of immutables in parameter
to this function. It could lead us to large lines of code but I think it is more
robust for the future of the code.

In my unit tests, I use to repeat code sometimes, because I find it clearer and
easier to maintain, tell me if you prefer me to factorize and mutualize some
parts in functinos.


gregorylegarec:
Sorry, too much commits. I'll check what gone wrong.

PR #906, opened 2016-07-29
Migrate Contact Store to Contact Reducer using Redux

gregorylegarec:
Please @aenario have a review, I d'like to have your opinion about differents
points:
* If I used Immutable correctly
* If my redux reducer is OK

Aslo, I spent some time refactoring the part of the code in charge of mapping
raw contacts to immutable contacts and raw contacts list onto an immubtable map.
I choose to create a dedicated library in libs/mappers, tell me if it is ok for
you.

Also, I choose to change a little bit the logic to have :

* a mapper function which map a raw contact to an immutable, without dealing
with the way they're will be mapped in an immutable map. It means that I add
a property `addresses` which keep all the email address of a contact. I also
kept the `address` property which store the contact primary email (it is also
used to to store the email used as index).
* a mapper function which map a raw contact list to an array of immutables
* a mapper function which return a mutator, to pass to
`Immutable#Map#withMutations` method. I choose to keep the raw contacts to
immutables logic outside, so we have to pass an array of immutables in parameter
to this function. It could lead us to large lines of code but I think it is more
robust for the future of the code.

In my unit tests, I use to repeat code sometimes, because I find it clearer and
easier to maintain, tell me if you prefer me to factorize and mutualize some
parts in functinos.


client/app/libs/mappers/contact.coffee
```diff
@@ -0,0 +1,72 @@
...
+        else
+            avatar = rawContact.datapoints?.find(_byName('avatar'))?.value
+
+        if avatar
+            contact = contact.set 'avatar', avatar
+
+        rawContact
+            .datapoints?.filter _byName 'email'
+            .forEach (datapoint) ->
+                if contact.has 'addresses'
```
aenario: you could simplify this part by using a Immutable.Set() for `addresses`

client/app/libs/mappers/contact.coffee
```diff
@@ -0,0 +1,72 @@
...
+
+        return contact
+
+
+    # Returns a list of Immutable contacts from a list of raw contacts
+    # rawContacts: an array of rawContacts
+    toImmutables: (rawContacts) ->
+        return unless rawContacts
+        rawContacts = [ rawContacts ] unless Array.isArray rawContacts
+        return rawContacts.map @toImmutable
```
aenario: How about making state.contacts a Map address -> contact instead, it will improve lookup performances

aenario: OK, thats the role of the mutator below, my bad

aenario: On the other end, it would be better to iterate contacts only once, so you could use `mapper.toImmutable` directly in toMapMutator.

aenario:
To ease up review, it would be nice to keep files name (contact_store.spec.js -> contact_reducer.spec.js) so we can see the diff on github. We can make all the renamings in a separate step at end of sprint.

gregorylegarec:
Ok for keeping files names. Thing is I almost rewrote the whole test file from scratch.

But I see your point and I will let the same name in the next PR. :)

client/app/libs/mappers/contact.coffee
gregorylegarec: Done, I totally removed `toInstances` which became useless.

client/test/contact_mapper.spec.js
```diff
@@ -0,0 +1,173 @@
+
+'use strict';
+const assert = require('chai').assert;
+const Immutable = require('immutable');
+
+const contactMapper = require('../app/libs/mappers/contact')
+
+const fixtures = {
```
misstick: I think that we should move this into `test/fixtures/contact.js` because contact should also be tested into messages. cc @aenario, @m4dz

PR #908, opened 2016-08-04
Migrate Settings Store to redux

gregorylegarec:
Thanks @anaerio to review this PR.

As we discussed, I created a pure getter for Settings.

I hesitated on the way to build the settings State, between using a single Map for the whole state or using a property.

```coffeescript
DEFAULT_STATE = Immutable.Map()
#or
DEFAULT_STATE =
  settings: Immutable.Map()
```

I kept the second solution to be consistent with the other migrations, but it forces us to call :

```coffeescript
reduxStore.getState().settings.settings
```

### DIFF
--- a/client/app/reducers/settings.coffee
+++ b/client/app/reducers/settings.coffee
@@ -2,14 +2,13 @@ Immutable = require 'immutable'
 
 {ActionTypes} = require '../constants/app_constants'
 
-DEFAULTSTATE =
-    settings: Immutable.Map()
+DEFAULTSTATE = Immutable.Map()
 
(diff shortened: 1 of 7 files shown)


client/app/reducers/settings.coffee
```diff
@@ -0,0 +1,15 @@
+Immutable = require 'immutable'
+
+{ActionTypes} = require '../constants/app_constants'
+
+DEFAULTSTATE =
+    settings: Immutable.Map()
```
aenario: I agree with you, let's keep the state as simple as possible. (no settings.settings)

